PR #5, opened 2016-06-15
GUACAMOLE-52: Document changes to history table schema

mike-jumper:


src/chapters/jdbc-auth.xml
```diff
@@ -839,15 +839,32 @@ INSERT INTO guacamole_connection_parameter VALUES (<replaceable>1</replaceable>,
...
-                                connection being used.</para>
+                                connection being used. If the connection associated with the history
+                                record no longer exists, this will be
+                                <constant>NULL</constant>.</para>
+                        </listitem>
+                    </varlistentry>
+                    <varlistentry>
+                        <term><property>connection_name</property></term>
+                        <listitem>
+                            <para>The unique name associated with the connection at the time that it
```
jmuehlner: Is the connection name actually unique?

mike-jumper: Yup. It isn't necessarily unique in the API, but it is in the schema, and that's enforced by a constraint.

jmuehlner: Well it's not globally unique right? Only within the parent connection group.

mike-jumper: True.

Very well. I shall document it ... *correctly*.

PR #6, opened 2016-08-26
GUACAMOLE-87: Bump guacamole-manual version numbers to 0.9.10-incubating.

jmuehlner:


src/chapters/jdbc-auth.xml
```diff
@@ -170,7 +170,7 @@ Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.
                     script located within the <filename>upgrade/</filename> directory:</para>
                 <informalexample>
                     <screen><prompt>$</prompt> <userinput>ls schema/upgrade/</userinput>
-<computeroutput>upgrade-pre-0.8.2.sql upgrade-pre-0.9.7.sql upgrade-pre-0.9.9.sql
+<computeroutput>upgrade-pre-0.8.2.sql upgrade-pre-0.9.7.sql upgrade-pre-0.9.10-incubating.sql
```
mike-jumper: Why do you hate the pre 0.9.9 upgrade script?

src/chapters/jdbc-auth.xml
```diff
@@ -201,11 +201,11 @@ upgrade-pre-0.9.6.sql upgrade-pre-0.9.8.sql</computeroutput>
 <prompt>Enter password:</prompt> <userinput><replaceable>password</replaceable></userinput>
 <prompt>$</prompt></screen>
                 </informalexample>
-                <para>If you are upgrading from a version of Guacamole before 0.9.9, schema changes
+                <para>If you are upgrading from a version of Guacamole before 0.9.10-incubating, schema changes
```
mike-jumper: Sorry, but we need to add a *new* section for the 0.9.10-incubating upgrade script. Changing the 0.9.9 section to instead point to 0.9.10-incubating makes it wrong. :(

src/chapters/jdbc-auth.xml
mike-jumper: Also, I believe the upgrade script is `upgrade-pre-0.9.10.sql` (and probably should stay that way for the sake of automation).

PR #33, opened 2017-02-13
GUACAMOLE-197: RADIUS Authentication Extension

necouchman:
I went ahead and took a stab at creating the manual page for the RADIUS authentication module.  Not sure exactly what style guidelines are for these XML files, or how to generate the HTML and/or PDF from this repo, but here's the initial commit.

### DIFF
--- a/src/chapters/radius-auth.xml
+++ b/src/chapters/radius-auth.xml
@@ -159,7 +159,7 @@
                     <para>The combination certificate and private key pair to use for TLS-based
                         RADIUS protocols that require a client-side certificate.  This parameter
                         should specify the absolute path to the file.  By default the extension
-                        will look for a file called radius.pem in the GUACAMOLE_HOME directory.
+                        will look for a file called radius.key in the GUACAMOLE_HOME directory.
                         </para>
                 </listitem>
             </varlistentry>
--- a/src/chapters/radius-auth.xml
+++ b/src/chapters/radius-auth.xml
@@ -47,47 +47,54 @@
                     below.</para>
             </step>
         </procedure>
-        <section xml:id="guac-radius-config">
-            <title>Configuring Guacamole for RADIUS authentication</title>
-            <indexterm>
-                <primary>configuring RADIUS authentication</primary>
-            </indexterm>
-            <indexterm>
-                <primary>RADIUS authentication</primary>
-                <secondary>configuration</secondary>
-            </indexterm>
-            <para>This extension provides several configuration properties in order 
-                to communicate properly with the RADIUS server to which it needs to authenticate.  It is 
-                important that you know several key pieces of information about the RADIUS server - 
-                at a minimum, the server name or IP, the Authentication port, the authentication 
-                protocol in use by the server, and the shared secret for the RADIUS client.  If you 
-                are responsible for the RADIUS server, you'll need to properly configure these items 
-                to get Guacamole to authenticate properly.  If you're not responsible for the RADIUS 
-                server you will need to work with the administrator to get all of the necessary 
-                configuration items for the server.  These items will need to be configured in the 
-                <link linkend="initial-setup"><filename>guacamole.properties</filename></link> 
-                file.</para>
-            <variablelist>
-                <varlistentry>
-                    <term><property>radius-server</property></term>
-                    <listitem>
-                        <para>The RADIUS server to authenticate against.  If not specified, 
-                            localhost will be used.</para>
-                    </listitem>
-                    <term><property>radius-auth-port</property></term>
-                    <listitem>
-                        <para>The RADIUS authentication port on which the RADIUS service is
-                            is listening.  If not specified, the default of 1812 will be
-                            used.</para>
-                    </listitem>
-                    <term><property>radius-shared-secret</property></term>
-                    <listitem>
-                        <para>The shared secret to use when talking to the RADIUS server.  This
+    </section>
+    <section xml:id="guac-radius-config">
+        <title>Configuring Guacamole for RADIUS authentication</title>
+        <indexterm>
+            <primary>configuring RADIUS authentication</primary>
+        </indexterm>
+        <indexterm>
+            <primary>RADIUS authentication</primary>
+            <secondary>configuration</secondary>
+        </indexterm>
+        <para>This extension provides several configuration properties in order 
+            to communicate properly with the RADIUS server to which it needs to authenticate.  It is 
+            important that you know several key pieces of information about the RADIUS server - 
+            at a minimum, the server name or IP, the Authentication port, the authentication 
+            protocol in use by the server, and the shared secret for the RADIUS client.  If you 
+            are responsible for the RADIUS server, you'll need to properly configure these items 
+            to get Guacamole to authenticate properly.  If you're not responsible for the RADIUS 
+            server you will need to work with the administrator to get all of the necessary 
+            configuration items for the server.  These items will need to be configured in the 
+            <link linkend="initial-setup"><filename>guacamole.properties</filename></link> 
+            file.</para>
+        <variablelist>
+            <varlistentry>
+                <term><property>radius-server</property></term>
+                <listitem>
+                    <para>The RADIUS server to authenticate against.  If not specified, 
+                        localhost will be used.</para>
+                </listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><property>radius-auth-port</property></term>
+                <listitem>
+                    <para>The RADIUS authentication port on which the RADIUS service is
+                        is listening.  If not specified, the default of 1812 will be
+                        used.</para>
+                </listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><property>radius-shared-secret</property></term>
+                <listitem>
+                    <para>The shared secret to use when talking to the RADIUS server.  This
                         parameter is required and defaults to null if not specified.</para>
-                    </listitem>
-                    <term><property>radius-auth-protocol</property></term>
-                    <listitem>
-                        <para>The authentication protocol to use when talking to the RADIUS server.
+                </listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><property>radius-auth-protocol</property></term>
+                <listitem>
+                    <para>The authentication protocol to use when talking to the RADIUS server.
                         This parameter is required for the extension to operate.  Supported
                         values are: pap, chap, mschapv1, mschapv2, eap-md5, eap-tls, and eap-ttls.
                         Support for PEAP is implemented inside the extension, but, due to a regression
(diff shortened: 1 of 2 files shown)


mike-jumper:
Thank you thank you thank you for taking it upon yourself to write some docs. ;)

> Not sure exactly what style guidelines are for these XML files, or how to generate the HTML and/or PDF from this repo, but here's the initial commit.

Instructions for building the manual can be found in the `README`: https://github.com/apache/incubator-guacamole-manual/blob/master/README

The repository contains a `Makefile` which takes care of the build, assuming the Docbook-related dependencies are present. Since the location of the Docbook XSL files varies widely, the `Makefile` requires that this be specified via an environment variable.

necouchman:
Got it...documentation should be pretty much written, and it seems to compile and look okay.  There's an odd blank page at the end that I'm not sure why it's generating, and doesn't seem to be present on the other documentation pieces, but I can't track down the exact source of that problem.

mike-jumper:
You can just ignore the generated PDF. The important thing to check is that the HTML looks proper, as that is what gets uploaded to the website come release time.

That PDF actually hasn't generated correctly for quite some time, since many of the code examples overflow the page boundaries... Honestly, the project may be better off if we switched the manual to a strict HTML-only build. The PDF is from back in the day when we wanted to publish print copies of the manual.

necouchman:
Okay, I think this is good to go - HTML looks okay to me.

jmuehlner:
@necouchman  Oops! Looks like there's some merge conflicts now!

necouchman:
@jmuehlner Thanks...rebased on master, should be good (for now)!

necouchman:
@mike-jumper @jmuehlner 
Documentation for recently-merged RADIUS module - I updated a couple of things, including adding more explicit build instructions for the module since it requires the extra Maven profile.

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,240 @@
...
+                <listitem>
+                    <para>The type of file store used for the certificate authority.  If not
+                        specified this defaults to pem.</para>
+                </listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><property>ca-key-password</property></term>
+                <listitem>
+                    <para>The password used to protect the certificate authority store, if
+                        any.  Default to null.</para>
```
mike-jumper: "Default to null" will not have meaning to users of the extension. The default behavior of this property needs to be described in terms not specific to the internals/code.

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,240 @@
...
+                </listitem>
+            </varlistentry>
+            <varlistentry>
+                 <term><property>radius-trust-all</property></term>
+                 <listitem>
+                     <para>Trust all server certificates without verifying against a CA file.
```
mike-jumper: But what value should be specified here? What is the default behavior if omitted?

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,240 @@
...
+                    <para>The password of the private key specified in the 
+                        <property>radius-key-file</property> parameter.  Defaults to null if not specified.</para>
+                </listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><property>ca-key-file</property></term>
+                <listitem>
+                    <para>The file that stores the certificate authority certificates for the
+                        connection to the RADIUS server.  Defaults to no file - if specified, the
+                        file must be present in the GUACAMOLE_HOME directory.</para>
```
mike-jumper: Does this mean that the entire path specified here is interpreted relative to `GUACAMOLE_HOME`? Or that this must be the name of a file which is present directly within `GUACAMOLE_HOME`?

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,240 @@
...
+                     <para>The type of the key file specified by the <property>radius-key-file</property> 
+                        parameter.  If not specified, this defaults to pkcs12, the default used by 
+                        the JRadius library.</para>
+                 </listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><property>radius-key-password</property></term>
+                <listitem>
+                    <para>The password of the private key specified in the 
+                        <property>radius-key-file</property> parameter.  Defaults to null if not specified.</para>
```
mike-jumper: "Defaults to null" only has useful meaning within the internals/code of the extension. This should be rephrased to something which makes sense from the user's point of view.

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,240 @@
...
+                    <para>The RADIUS authentication port on which the RADIUS service is
+                        is listening.  If not specified, the default of 1812 will be
+                        used.</para>
+                </listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><property>radius-shared-secret</property></term>
+                <listitem>
+                    <para>The shared secret to use when talking to the RADIUS server.  This
+                        parameter is required and defaults to null if not specified.</para>
```
mike-jumper: "Defaults to null" only has useful meaning within the internals/code of the extension. This should be rephrased to something which makes sense from the user's point of view.

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,240 @@
...
+                    <para>The timeout for a RADIUS connection in seconds.  The default is
+                        60 seconds.</para>
+                </listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><property>radius-eap-ttls-inner-protocol</property></term>
+                <listitem>
+                    <para>When EAP-TTLS is used, this parameter specifies the inner (tunneled)
+                        protocol to use talking to the RADIUS server.  It is required when the
+                        <property>radius-auth-protocol</property> parameter is set to eap-ttls.</para>
```
mike-jumper: What is the behavior of this parameter if `radius-auth-protocol` is *not* set to `eap-ttls`? If it has no effect, we should note that here.

src/chapters/radius-auth.xml
mike-jumper: In fact ... it might be easier to conceptualize for users, easier to document, and cleaner to code if things are instead done with a `FileGuacamoleProperty`:

https://github.com/apache/guacamole-client/blob/798349e2310d3ba90e3a1d5fa276c2c8efc19c82/guacamole-ext/src/main/java/org/apache/guacamole/properties/FileGuacamoleProperty.java

That would tie the type of the property to `File`, rather than having to interpret the `String` manually, and would result in the path being interpreted as an absolute path. Relying on a file with a standardized filename within `GUACAMOLE_HOME` can then be left as the default.

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,240 @@
...
+            </varlistentry>
+            <varlistentry>
+                <term><property>radius-key-password</property></term>
+                <listitem>
+                    <para>The password of the private key specified in the 
+                        <property>radius-key-file</property> parameter.  Defaults to null if not specified.</para>
+                </listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><property>ca-key-file</property></term>
```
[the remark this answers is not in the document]
necouchman: These are actually correct in the extension, I just forgot to change them in the docs.  Will update it.

src/chapters/radius-auth.xml
necouchman: Okay, I will reword documentation, and then rework the extension to go this route.

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,240 @@
...
+        <primary>RADIUS Authentication</primary>
+    </indexterm>
+    <para>Guacamole supports delegating authentication to a RADIUS service, such as FreeRADIUS, to
+        validate username and password combinations, and to support multi-factor authentication.  This
+        authentication method must be layered on top of some other authentication extension, such as
+        those available from the main project website, in order to provide access to actual
+        connections.</para>
+    <section xml:id="radius-downloading">
+        <title>Downloading the RADIUS authentication extension</title>
+        <para>Due to licensing limitations on dependencies for the RADIUS extension, the extension
```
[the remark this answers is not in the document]
necouchman: Let me know if this looks okay, now, or if you want me to mention JRadius specifically, etc.

src/chapters/radius-auth.xml
necouchman: Should be better.

src/chapters/radius-auth.xml
necouchman: Fixed.

src/chapters/radius-auth.xml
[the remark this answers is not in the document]
necouchman: Updated.

src/chapters/radius-auth.xml
necouchman: Documentation reworded, and submitted a new PR for the client to modify the extension to use `FileGuacamoleProperty`.

src/chapters/radius-auth.xml
necouchman: Fixed.

src/chapters/radius-auth.xml
necouchman: Fixed.

src/chapters/radius-auth.xml
necouchman: Fixed.

necouchman:
Okay, cleaned this up with your suggestions.

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,251 @@
...
+                <listitem>
+                    <para>The password used to protect the certificate authority store, if
+                        any.  If unspecified the extension will attempt to read the CA
+                        store without any password..</para>
+                </listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><property>radius-trust-all</property></term>
+                <listitem>
+                    <para>A boolean value indicating whether or not the RADIUS extension
```
mike-jumper: Similar to the problem with telling users the default is "null", documenting that the property is a boolean value is not necessarily sufficient to inform users that the property can be set to "true" or "false". The default is nicely documented here, but the fact that this should be set to "true" to trust all certificates needs to be explicit.

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,251 @@
...
+                    <para>The absolute path to the file that stores the certificate authority
+                        certificates for encrypted connections to the RADIUS server.  By default
+                        a file with the name ca.crt in the GUACAMOLE_HOME directory will be used.
+                        </para>
+                </listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><property>radius-ca-type</property></term>
+                <listitem>
+                    <para>The type of file store used for the certificate authority.  If not
```
mike-jumper: Do you mean key store?

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,251 @@
...
+                        specified this defaults to pem.</para>
```
mike-jumper: What are the legal/supported values?

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,251 @@
...
+                        will look for a file called radius.key in the GUACAMOLE_HOME directory.
+                        </para>
+                </listitem>
+            </varlistentry>
+            <varlistentry>
+                 <term><property>radius-key-type</property></term>
+                 <listitem>
+                     <para>The type of the key file specified by the <property>radius-key-file</property> 
+                        parameter.  If not specified, this defaults to pkcs12, the default used by 
+                        the JRadius library.</para>
```
mike-jumper: What are the legal/supported values?

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,251 @@
...
+                    <para>The type of file store used for the certificate authority.  If not
+                        specified this defaults to pem.</para>
+                </listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><property>radius-ca-password</property></term>
+                <listitem>
+                    <para>The password used to protect the certificate authority store, if
+                        any.  If unspecified the extension will attempt to read the CA
+                        store without any password..</para>
```
mike-jumper: There are two periods after the sentence here.

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,251 @@
...
+                    should trust all certificates or verify them against known good
+                    certificate authorities.  By default this is false, and certificates
+                    will validated.</para>
+                </listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><property>radius-retries</property></term>
+                <listitem>
+                    <para>The number of times the client will retry the connection to the 
+                        RADIUS server and not receive a response before giving up.  The default
```
mike-jumper: For clarity, rather than simply state the effective default value, we should also state the default behavior. This is done nicely for `radius-trust-all` above:

> By default this is false, and certificates will validated.

but here is less complete:

> The default is 5.

Something like "By default, the client will retry the connection at most 5 times." (or similar) would satisfy this need.

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,251 @@
...
+                        is 5.</para>
+                </listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><property>radius-timeout</property></term>
+                <listitem>
+                    <para>The timeout for a RADIUS connection in seconds.  The default is
+                        60 seconds.</para>
```
mike-jumper: Though not as unclear as with `radius-retries`, I think the same should be done here for the sake of clarity: describe the default behavior, not just the default value.

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,251 @@
...
+                </listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><property>radius-eap-ttls-inner-protocol</property></term>
+                <listitem>
+                    <para>When EAP-TTLS is used, this parameter specifies the inner (tunneled)
+                        protocol to use talking to the RADIUS server.  It is required when the
```
mike-jumper: What are the legal/supported values?

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,251 @@
...
+            <primary>configuring RADIUS authentication</primary>
+        </indexterm>
+        <indexterm>
+            <primary>RADIUS authentication</primary>
+            <secondary>configuration</secondary>
+        </indexterm>
+        <para>This extension provides several configuration properties in order 
+            to communicate properly with the RADIUS server to which it needs to authenticate.  It is 
+            important that you know several key pieces of information about the RADIUS server - 
+            at a minimum, the server name or IP, the Authentication port, the authentication
```
mike-jumper: Why is "authentication" capitalized here?

src/chapters/radius-auth.xml
```diff
@@ -0,0 +1,251 @@
...
+            dependency, the Guacamole project cannot distribute binary versions of the RADIUS extension.
+            If you want to use this extension you will need to build the code - or at least the RADIUS
+            extension yourself.  Build instructions can be found in the section
+            <xref linkend="installing-guacamole"/>.</para>
+    </section>
+    <section xml:id="installing-radius-auth">
+        <title>Installing RADIUS authentication</title>
+        <para>The RADIUS extension must be explicitly enabled during build time in order to generate
+            the binaries and resulting JAR file.  This is done by adding the flag <option>-Plgpl-extensions</option>
+            to the MAVEN command line during the build, and should result in the output below:</para>
```
mike-jumper: Maven is not an acronym (see http://maven.apache.org/).

necouchman: Fixed.

src/chapters/radius-auth.xml
necouchman: Fixed.

src/chapters/radius-auth.xml
necouchman: Added.

src/chapters/radius-auth.xml
necouchman: Fixed.

src/chapters/radius-auth.xml
necouchman: Added.

src/chapters/radius-auth.xml
necouchman: Fixed.

src/chapters/radius-auth.xml
necouchman: Fixed.

src/chapters/radius-auth.xml
necouchman: Fixed.

src/chapters/radius-auth.xml
necouchman: Added.

necouchman:
Next round...

PR #38, opened 2017-03-18
GUACAMOLE-244: Configuration of Dereferencing Aliases

necouchman:
This adds documentation for the single parameter added to the client that allows for configuring when the LDAP connection follows (dereferences) aliases.

### DIFF
--- a/src/chapters/ldap-auth.xml
+++ b/src/chapters/ldap-auth.xml
@@ -391,7 +391,8 @@ dn: cn={4}guacConfigGroup,cn=schema,cn=config
                             as it searches the tree.  Possible values for this property are "never" (the default)
                             so that aliases will never be followed, "searching" to dereference during search operations
                             after the base object is located, "finding" to dereference in order to locate the
-                            search base, but not during the actual search, and "always" to always dereference aliases.
+                            search base, but not during the actual search, and "always" to always dereference
+                            aliases.</para>
                     </listitem>
                 </varlistentry>
             </variablelist>
--- a/src/chapters/ldap-auth.xml
+++ b/src/chapters/ldap-auth.xml
@@ -388,11 +388,10 @@ dn: cn={4}guacConfigGroup,cn=schema,cn=config
                     <term><property>ldap-dereference-aliases</property></term>
                     <listitem>
                         <para>Controls whether or not the LDAP connection follows (dereferences) aliases
-                            as it searches the tree.  There are four possible values: never (the default),
-                            finding, searching, and always.  Never means aliases will never be followed;
-                            searching means they will be dereferenced during search, but not when finding
-                            the start of the search; finding means only when finding the start of the search,
-                            but not during the search; always means they will always be followed.</para>
+                            as it searches the tree.  Poossible values for this property are "never" (the default)
+                            so that aliases will never be followed, "searching" to dereference during search operations
+                            after the base object is located, "finding" to dereference in order to locate the
+                            search base, but not during the actual search, and "always" to always dereference aliases.
                     </listitem>
                 </varlistentry>
             </variablelist>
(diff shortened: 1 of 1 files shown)


src/chapters/ldap-auth.xml
```diff
@@ -384,6 +384,17 @@ dn: cn={4}guacConfigGroup,cn=schema,cn=config
                                 have no effect on Guacamole configurations.</emphasis></para>
                     </listitem>
                 </varlistentry>
+                <varlistentry>
+                    <term><property>ldap-dereference-aliases</property></term>
+                    <listitem>
+                        <para>Controls whether or not the LDAP connection follows (dereferences) aliases
+                            as it searches the tree.  There are four possible values: never (the default),
```
mike-jumper: The values themselves need to be in quotes, so it's clear that their literal values must be specified verbatim. This is the convention that we follow elsewhere. For example, a few sections above:

    <para>The encryption mechanism that Guacamole should use when communicating
        with your LDAP server. Legal values are "none" for unencrypted LDAP,
        "ssl" for LDAP over SSL/TLS (commonly known as LDAPS), or "starttls" for
        STARTTLS. If omitted, encryption will not be used.</para>

src/chapters/ldap-auth.xml
```diff
@@ -384,6 +384,17 @@ dn: cn={4}guacConfigGroup,cn=schema,cn=config
...
+                            finding, searching, and always.  Never means aliases will never be followed;
```
mike-jumper: Similar to the above, starting this sentence with "Never means ..." (uppercase N) could lead to confusion when specifying the `ldap-dereference-aliases` property, as the values are strictly interpreted. Things need to be rephrased/reorganized somehow such that the value itself can be written lowercase, exactly as it would appear in `guacamole.properties`, and in quotes so it's clear that we're referring to a strict literal value.

src/chapters/ldap-auth.xml
necouchman: Fixed.

src/chapters/ldap-auth.xml
necouchman: Reworded.

src/chapters/ldap-auth.xml
```diff
@@ -388,11 +388,10 @@ dn: cn={4}guacConfigGroup,cn=schema,cn=config
                     <term><property>ldap-dereference-aliases</property></term>
                     <listitem>
                         <para>Controls whether or not the LDAP connection follows (dereferences) aliases
-                            as it searches the tree.  There are four possible values: never (the default),
-                            finding, searching, and always.  Never means aliases will never be followed;
-                            searching means they will be dereferenced during search, but not when finding
-                            the start of the search; finding means only when finding the start of the search,
-                            but not during the search; always means they will always be followed.</para>
+                            as it searches the tree.  Poossible values for this property are "never" (the default)
```
mike-jumper: "Poossible", you say. 💩

necouchman: Fixed.

src/chapters/ldap-auth.xml
```diff
@@ -384,6 +384,16 @@ dn: cn={4}guacConfigGroup,cn=schema,cn=config
...
                     </listitem>
                 </varlistentry>
+                <varlistentry>
+                    <term><property>ldap-dereference-aliases</property></term>
+                    <listitem>
+                        <para>Controls whether or not the LDAP connection follows (dereferences) aliases
+                            as it searches the tree.  Possible values for this property are "never" (the default)
+                            so that aliases will never be followed, "searching" to dereference during search operations
+                            after the base object is located, "finding" to dereference in order to locate the
+                            search base, but not during the actual search, and "always" to always dereference aliases.
```
mike-jumper: Your `<para>` is missing a corresponding end tag, which is causing the manual build to fail with these changes present:

```
$ make
cd src; xsltproc -o ../html/ --xinclude site.xslt gug.xml
chapters/ldap-auth.xml:395: parser error : Opening and ending tag mismatch: para line 390 and listitem
                    </listitem>
                               ^
```

necouchman: Fixed, and built it to make sure.

PR #39, opened 2017-03-21
GUACAMOLE-101: Documentation for LDAP User Filtering

necouchman:
Added documentation for the two properties added for supporting filtering of users and connections in LDAP.

### DIFF
--- a/src/chapters/ldap-auth.xml
+++ b/src/chapters/ldap-auth.xml
@@ -404,15 +404,6 @@ dn: cn={4}guacConfigGroup,cn=schema,cn=config
                             </emphasis></para>
                     </listitem>
                 </varlistentry>
-                <varlistentry>
-                    <term><property>ldap-connection-search-filter</property></term>
-                    <listitem>
-                        <para>The search filter used to query the LDAP tree for connections that
-                            can be accessed from Guacamole.  <emphasis>If this property is omitted
-                            the default of (objectClass=guacConfigGroup) will be used.
-                            </emphasis></para>
-                    </listitem>
-                </varlistentry>
             </variablelist>
             <para>Again, even if the defaults are sufficient for the other properties, <emphasis>you
                     must still specify the <property>ldap-user-base-dn</property>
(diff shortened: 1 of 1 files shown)


necouchman:
Removed documentation related to connection filtering, as that has been removed from the code.

src/chapters/ldap-auth.xml
```diff
@@ -384,6 +384,15 @@ dn: cn={4}guacConfigGroup,cn=schema,cn=config
                                 have no effect on Guacamole configurations.</emphasis></para>
                     </listitem>
                 </varlistentry>
+                <varlistentry>
+                    <term><property>ldap-user-search-filter</property></term>
+                    <listitem>
+                        <para>The search filter used to query the LDAP tree for users that
+                            can log into and be granted privileges in Guacamole.  <emphasis>If
+                            this property is omitted the default of (objectClass=*) will be used.
```
mike-jumper: Please place `(objectClass=*)` within quotes, so it's clear that it's a literal value.

necouchman: Fixed.

mike-jumper:
Please rebase - things conflict against current `master`.

necouchman:
Rebased.

src/chapters/ldap-auth.xml
```diff
@@ -395,6 +395,15 @@ dn: cn={4}guacConfigGroup,cn=schema,cn=config
                             aliases.</para>
                     </listitem>
                 </varlistentry>
+                <varlistentry>
+                    <term><property>ldap-user-search-filter</property></term>
```
mike-jumper: OK - I built the manual from these changes and noticed that the organization of this section is slightly off. There's a section in the middle of this area which is devoted to users and search:

> ldap-user-search-stuff

but the new `ldap-user-search-filter` property documentation is simply appended to the end of the section among other unrelated properties:

> ldap-user-search-filter-section

Perhaps this should be moved up a bit?

necouchman: Yeah, that makes sense.  I moved it just below the ldap-username-attribute.

PR #46, opened 2017-05-31
GUACAMOLE-203: Update SSH documentation to include keepalive interval

necouchman:
Adding the new server-alive-interval parameter to the documentation for SSH connections.

### DIFF
--- a/src/chapters/configuring.xml
+++ b/src/chapters/configuring.xml
@@ -2587,6 +2587,24 @@ ed272546-87bd-4db9-acba-e36e1a9ca20a
                                         the default of 22 will be used.</para>
                                 </entry>
                             </row>
+                            <row>
+                                <entry><parameter>server-alive-interval</parameter></entry>
+                                <entry>
+                                    <para>
+                                        <indexterm>
+                                            <primary>SSH</primary>
+                                            <secondary>server-alive-interval</secondary>
+                                        </indexterm>
+                                        By default the SSH client does not send keepalive requests
+                                        to the server.  This parameter allows you to configure the
+                                        the interval (in seconds) at which the client connection
+                                        sends keepalive packets to the server.  The default is 0,
+                                        which disables sending the packets.  A value of 1 is
+                                        automatically increased to 2.  Negative values are set to
+                                        0, disabling sending keepalives.
+                                    </para>
+                                </entry>
+                            </row>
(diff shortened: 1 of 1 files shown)


necouchman:
Added documentation for the SFTP parameters.

src/chapters/configuring.xml
```diff
@@ -2772,6 +2794,24 @@ ed272546-87bd-4db9-acba-e36e1a9ca20a
                                         session will use the user's default shell.</para>
                                 </entry>
                             </row>
+                            <row>
+                                <entry><parameter>server-alive-interval</parameter></entry>
```
mike-jumper: I'm not sure it makes sense to include the keepalive interval within a new "environment settings" section. Since the keepalive interval deals with the SSH connection itself, what about the existing "network parameters" section?

http://guacamole.incubator.apache.org/doc/gug/configuring-guacamole.html#ssh-network-parameters

IMHO, the keepalive interval is by definition a network parameter.

src/chapters/configuring.xml
```diff
@@ -2772,6 +2794,24 @@ ed272546-87bd-4db9-acba-e36e1a9ca20a
...
+                                    <para>
+                                        <indexterm>
+                                            <primary>SSH</primary>
+                                            <secondary>server-alive-interval</secondary>
+                                        </indexterm>
+                                        By default the SSH client does not send keepalive requests
+                                        to the server.  This parameter allows you to configure the
+                                        the interval (in seconds) at which the client connection
+                                        sends keepalive packets to the server.  The default is 0,
+                                        which disables sending the packets.  A value of 1 is
```
mike-jumper: > ... A value of 1 is automatically increased to 2.  Negative values are set to 0, disabling sending keepalives.

The "default is zero, one becomes two, negative becomes zero" description seems convoluted/confusing. It makes sense in the code, but this description borders on being pseudocode.

Wouldn't it be easier to understand as simply being disabled by default, with the minimum legal value being 2?

src/chapters/configuring.xml
necouchman: Moved.

src/chapters/configuring.xml
necouchman: Okay, should be cleaned up, now.

PR #48, opened 2017-06-04
GUACAMOLE-292: Document new user profile columns of guacamole_user.

mike-jumper:


mike-jumper:
**NOTE:** Intended for `staging/0.9.13-incubating`.

src/chapters/jdbc-auth.xml
```diff
@@ -805,6 +805,48 @@ postgresql-user-required: true</programlisting>
                             zones.</para>
                     </listitem>
                 </varlistentry>
+                <varlistentry>
+                    <term><property>full_name</property></term>
+                    <listitem>
+                        <para>The user's full name. Unlike the username, this name need not be
```
jmuehlner: Not that I think it's super needed, but why does this paragraph not have a thing about setting the column to NULL like the others?

mike-jumper: I'll add a thing.

mike-jumper: There - hopefully that adds some clarity.

jmuehlner: #CONSISTENCY

PR #49, opened 2017-06-04
GUACAMOLE-189: Document guacd override columns of guacamole_connection table.

mike-jumper:


mike-jumper:
**NOTE:** Intended for `staging/0.9.13-incubating`.

PR #55, opened 2017-08-01
GUACAMOLE-357: Update Documentation for Ubuntu Specific Library

MysticRyuujin:
On Ubuntu the libjpeg62-turbo-dev package is named libjpeg-turbo8-dev, the instructions don't make that clear, I'd like to update the documentation to reflect that distinction.

### DIFF
--- a/src/chapters/installing.xml
+++ b/src/chapters/installing.xml
@@ -136,8 +136,12 @@
                                         <colspec colname="c2" colnum="2" colwidth="1.0*"/>
                                         <tbody>
                                             <row>
-                                                <entry>Debian / Ubuntu package</entry>
+                                                <entry>Debian package</entry>
                                                 <entry><package>libjpeg62-turbo-dev</package></entry>
+                                            </row>
+                                            <row>
+                                                <entry>Ubuntu package</entry>
+                                                <entry><package>libjpeg-turbo8-dev</package></entry>
                                             </row>
                                             <row>
                                                 <entry>Fedora / CentOS / RHEL package</entry>


src/chapters/installing.xml
```diff
@@ -136,10 +136,14 @@
                                         <colspec colname="c2" colnum="2" colwidth="1.0*"/>
                                         <tbody>
                                             <row>
-                                                <entry>Debian / Ubuntu package</entry>
+                                                <entry>Debian package</entry>
                                                 <entry><package>libjpeg62-turbo-dev</package></entry>
                                             </row>
                                             <row>
+                                                <entry>Ubuntu package</entry>
+                                                <entry><package>libjpeg-turbo8-dev</package></entry>
```
mike-jumper: Ubuntu uses Debian as the basis of each of their releases, right? Are you sure that Debian vs. Ubuntu is the dividing line between `libjpeg62-turbo-dev` and `libjpeg-turbo8-dev`?

MysticRyuujin:
So I've been using/building guacd from source since Ubuntu 14.x branches (14,15, and 16) and it's never worked with "libjpeg62-turbo-dev" it's a non existent package for Ubuntu. I've only ever made it work with "libjpeg-turbo8-dev" OR the non-turbo version which for whatever reason is still named "libjpeg62-dev"

If you just do a Google search for "Ubuntu libjpeg62-turbo-dev" you'll see what I'm talking about.

I cannot speak for all other Debian based distros because I've only used Ubuntu but with the popularity of Ubuntu I think it makes sense to put it in the manual.

mike-jumper:
Checking the Debian repositories, I can confirm that `libjpeg62-turbo-dev` is the only correct package there, so that's good. It's still that package, even on the latest unstable:

https://packages.debian.org/source/sid/libjpeg-turbo

As for Ubuntu, again, it looks like you're completely correct - even the oldest non-EOL version of Ubuntu (14.04) is on `libjpeg-turbo8-dev`:

https://packages.ubuntu.com/source/trusty/libjpeg-turbo

I'll go ahead and merge this.

PR #56, opened 2017-10-08
GUACAMOLE-409: Include a Dockerfile in guacamole-manual

ceharris:
This request adds a Dockerfile along with instructions for using it.

In working on the additions to the manual for the newly merged event listener support, I've found it to be helpful, and I suspect others will as well. Using the Dockerfile, it isn't necessary to install all of the Docbook dependencies in order to contribute to the documentation, build it, and see the resulting changes.


mike-jumper:
1. Please tag commit messages with the JIRA issue.
2. The three most recent commits look like they're related to the new event listener support, not to adding a Docker manual build. Wrong branch, I presume?

PR #58, opened 2017-10-09
WIP: GUACAMOLE-406: Add documentation for event listener support

ceharris:
I think this is at a point where some review would be helpful. Would appreciate feedback on the content itself (especially level of detail, illustrative value of the examples, etc), placement in the user manual structure, as well as any syntactic issues.

There is some repetition between this section and the section on custom authentication providers. In particular, the instructions for installing a custom extension are duplicated almost verbatim. Perhaps this would be better addressed in the chapter on guacamole-ext or elsewhere. 

Much of the skeleton for the example itself is also very similar to the authentication provider example, but I think this is probably to be expected and may not be a bad thing.


ceharris:
Not sure why I still have commits for GUACAMOLE-409 showing up here. Will rebase to clean that up.

PR #59, opened 2017-10-09
GUACAMOLE-406: Add documentation for event listener support

ceharris:
I think this is at a point where some review would be helpful. Would appreciate feedback on the content itself (especially level of detail, illustrative value of the examples, etc), placement in the user manual structure, as well as any syntactic issues.

There is some repetition between this section and the section on custom authentication providers. In particular, the instructions for installing a custom extension are duplicated almost verbatim. Perhaps this would be better addressed in the chapter on guacamole-ext or elsewhere.

Much of the skeleton for the example itself is also very similar to the authentication provider example, but I think this is probably to be expected and may not be a bad thing.

### DIFF
--- a/src/chapters/event-listeners.xml
+++ b/src/chapters/event-listeners.xml
@@ -7,21 +7,20 @@
         <primary>events</primary>
         <secondary>listeners</secondary>
     </indexterm>
-    <para>Guacamole provides supports the delivery of event notifications to custom extensions.
+    <para>Guacamole supports the delivery of event notifications to custom extensions.
         Developers can use listener extensions to integrate custom handling of events such as 
-        successful and failed authentications, and requests to connect and disconnect from desktop 
-        environments configured for Guacamole users.</para>
-    <para>A listener extension could be used, for example, to record successful or failed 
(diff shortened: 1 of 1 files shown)


ceharris:
In light of the fact that the new listener extension mechanism is slated for 0.9.14-incubating, it would be good to get this merged soon.

Can someone please review and comment?

mike-jumper:
Thanks for the ping, @ceharris - I don't think changes to the PR title get broadcast to the lists.

Regarding the changes, things generally look fine. There are API changes coming via #205, but this documentation doesn't look incompatible with those.

Since this is the documentation which corresponds to changes which are already part of 0.9.14-incubating ([GUACAMOLE-364](https://issues.apache.org/jira/browse/GUACAMOLE-364)), if you can then update the merge base from `master` to `0.9.14-incubating` this PR, I'd be glad to merge.

necouchman:
Looks great to me, as well, pending the change of the merge base from `master` to the `staging/0.9.14-incubating`.

ceharris:
Closing to switch to different source branch and merge base.

PR #63, opened 2017-10-28
GUACAMOLE-427: Fix documentation on load-balance-info parameter.

necouchman:
Quick fix.  Going to suggest this makes it into 0.9.14-incubating release.

### DIFF
--- a/src/chapters/configuring.xml
+++ b/src/chapters/configuring.xml
@@ -2220,7 +2220,7 @@ ed272546-87bd-4db9-acba-e36e1a9ca20a
                         </thead>
                         <tbody>
                             <row>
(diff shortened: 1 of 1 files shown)


mike-jumper:
As quick/tempting this fix may be, scope of 0.9.14-incubating has been finalized. Expanding the scope would turn the release into a moving target.

If you feel strongly this should go into 0.9.14-incubating, let's discuss expanding scope on the mailing lists. If not, let's keep the scope as it is.

necouchman:
No, don't feel that strongly about it.  Rebased it against master.

lorengordon:
Thanks @mike-jumper and @necouchman! You guys rock!

PR #66, opened 2017-11-30
GUACAMOLE-363: Add SQL Server documentation to database manual page.

necouchman:
This PR adds the documentation for the SQL Server JDBC extensions to the database authentication page.

Wasn't able to reopen the old PR, so doing a new one, here.

### DIFF
--- a/src/chapters/jdbc-auth.xml
+++ b/src/chapters/jdbc-auth.xml
@@ -121,21 +121,21 @@
                         mode of the module with various TDS-comptabile drivers such that it can be used with different
                         versions of SQL Server and even non-Microsoft SQL Server databases.  The following options are available
                         for the <property>sqlserver-driver</property> property:
-                        <variablelist>
-                            <varlistentry>
+                        <itemizedlist>
+                            <listitem>
                                 <constant>microsoft2005</constant>: The current Microsoft driver,
                                 supporting SQL Server 2005 and later.
-                            </varlistentry>
-                            <varlistentry>
+                            </listitem>
+                            <listitem>
                                 <constant>microsoft</constant>: The legacy SQL Server support.
-                            </varlistentry>
-                            <varlistentry>
+                            </listitem>
+                            <listitem>
                                 <constant>jtds</constant>: The open source JTDS driver.
-                            </varlistentry>
-                            <varlistentry>
+                            </listitem>
+                            <listitem>
                                 <constant>datadirect</constant>: The Progress Sybase driver.
-                            </varlistentry>
-                        </variablelist>
+                            </listitem>
+                        </itemizedlist>
                     </para>
                 </listitem>
             </varlistentry>
--- a/src/chapters/jdbc-auth.xml
+++ b/src/chapters/jdbc-auth.xml
@@ -124,19 +124,27 @@
                         <variablelist>
                             <varlistentry>
                                 <term>microsoft2005</term>
-                                <listitem>The current Microsoft driver, supporting SQL Server 2005 and later.</listitem>
+                                <listitem>
+                                    <para>The current Microsoft driver, supporting SQL Server 2005 and later.</para>
+                                </listitem>
                             </varlistentry>
                             <varlistentry>
                                 <term>microsoft</term>
-                                <listitem>The legacy SQL Server support.</listitem>
+                                <listitem>
+                                    <para>The legacy SQL Server support.</para>
+                                </listitem>
                             </varlistentry>
                             <varlistentry>
                                 <term>jtds</term>
-                                <listitem>The open source JTDS driver.</listitem>
+                                <listitem>
+                                    <para>The open source JTDS driver.</para>
+                                </listitem>
                             </varlistentry>
                             <varlistentry>
                                 <term>datadirect</term>
-                                <listitem>The Progress Sybase driver.</listitem>
+                                <listitem>
+                                    <para>The Progress Sybase driver.</para>
+                                </listitem>
                             </varlistentry>
                         </variablelist>
                     </para>
--- a/src/chapters/jdbc-auth.xml
+++ b/src/chapters/jdbc-auth.xml
@@ -108,13 +108,13 @@
                         SQL Server JDBC driver must be placed within the <filename>GUACAMOLE_HOME/lib</filename>
                         directory.</para>
                     <para><emphasis>The SQL Server JDBC driver is not included with the extension.</emphasis>  You
-                        must obtain the JDBC driver <filename>.jar</filename> yourselfand place it in the directory.
+                        must obtain the JDBC driver <filename>.jar</filename> yourself and place it in the directory.
                         Furthermore, the SQL Server authentication extension supports a number of TDS-compatible
                         drivers, so you must make sure the one you choose is supported by the extension, that the
                         extension is configured properly, and that the <filename>.jar</filename> is in the correct
-                        directory.  Microsoft's JDBC driver can be downloaded from <link
+                        directory.  Microsoft's JDBC driver can be downloaded from Microsoft's <link
                         xlink:href="https://docs.microsoft.com/en-us/sql/connect/sql-connection-libraries#anchor-20-drivers-relational-access">
-                        this page</link>.</para>
+                        SQL Connection Libraries</link> page.</para>
                 </listitem>
             </varlistentry>
         </variablelist>
@@ -324,21 +324,22 @@ Type "help" for help.
         <section xml:id="jdbc-auth-sqlserver">
             <title>SQL Server</title>
             <para>If using SQL Server, the database and schema must be created first.  The example below assumes
-                that you are running SQL Server on Linux, using the command-line tools to manage it.</para>
+                that you are running SQL Server on Linux, using the command-line tools to manage it, however, this
+                code can be run using any tool capable of running SQL against a SQL Server database.</para>
             <informalexample>
                 <screen><prompt>$</prompt> <userinput>/opt/mssql-tools/bin/sqlcmd -S localhost -U SA</userinput>
 <prompt>Password:</prompt> <userinput><replaceable>password</replaceable></userinput>
 <prompt>1></prompt> <userinput>CREATE DATABASE <replaceable>guacamole_db</replaceable>;</userinput>
 <prompt>2></prompt> <userinput>GO</userinput>
-<prompt>1></prompt> <userinput>CREATE LOGIN <replaceable>guacamole_user</replaceable> WITH PASSWORD = '<replaceable>S0me_Password</replaceable>';</userinput>
+<prompt>1></prompt> <userinput>CREATE LOGIN <replaceable>guacamole_user</replaceable> WITH PASSWORD = '<replaceable>some_password</replaceable>';</userinput>
 <prompt>2></prompt> <userinput>GO</userinput>
 <prompt>1></prompt> <userinput>USE <replaceable>guacamole_db</replaceable>;</userinput>
 <prompt>2></prompt> <userinput>GO</userinput>
 <prompt>1></prompt> <userinput>CREATE USER <replaceable>guacamole_user</replaceable>;</userinput>
 <prompt>2></prompt> <userinput>GO</userinput>
 <prompt>1></prompt> <userinput>ALTER ROLE db_datawriter ADD MEMBER <replaceable>guacamole_user</replaceable>;</userinput>
-<prompt>1></prompt> <userinput>ALTER ROLE db_datareader ADD MEMBER <replaceable>guacamole_user</replaceable>;</userinput>
-<prompt>2></prompt> <userinput>GO</userinput></screen>
+<prompt>2></prompt> <userinput>ALTER ROLE db_datareader ADD MEMBER <replaceable>guacamole_user</replaceable>;</userinput>
+<prompt>3></prompt> <userinput>GO</userinput></screen>
             </informalexample>
             <para>Once the database and user account is created, and the user associated with the database, you can use
                 the supplied scripts to load the schema into the database.  These scripts are included in the
@@ -346,13 +347,13 @@ Type "help" for help.
                 web site.</para>
             <informalexample>
                 <screen><prompt>$</prompt> <userinput>/opt/mssql-tools/bin/sqlcmd -S localhost -U <replaceable>guacamole_user</replaceable> -d <replaceable>guacamole_db</replaceable> -i schema/001-create-schema.sql</userinput>
-<prompt>Password:</prompt> <userinput><replaceable>S0me_Password</replaceable></userinput>
+<prompt>Password:</prompt> <userinput><replaceable>some_password</replaceable></userinput>
 <computeroutput>Rule bound to data type.
 The new rule has been bound to column(s) of the specified user data type.
 Rule bound to data type.
 The new rule has been bound to column(s) of the specified user data type.</computeroutput>
 <prompt>$</prompt> <userinput>/opt/mssql-tools/bin/sqlcmd -S localhost -U <replaceable>guacamole_user</replaceable> -d <replaceable>guacamole_db</replaceable> -i schema/002-create-admin-user.sql</userinput>
-<prompt>Password:</prompt> <userinput><replaceable>S0me_Password</replaceable></userinput>
+<prompt>Password:</prompt> <userinput><replaceable>some_password</replaceable></userinput>
 <computeroutput>
 (1 rows affected)
 
@@ -444,7 +445,7 @@ sqlserver-hostname: localhost
 sqlserver-port: 1433
 sqlserver-database: <replaceable>guacamole_db</replaceable>
 sqlserver-username: <replaceable>guacamole_user</replaceable>
-sqlserver-password: <replaceable>S0me_Password</replaceable>
+sqlserver-password: <replaceable>some_password</replaceable>
 sqlserver-driver: microsoft2005
                 </programlisting>
             </informalexample>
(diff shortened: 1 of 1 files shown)


src/chapters/jdbc-auth.xml
```diff
@@ -97,9 +97,29 @@
...
+                <listitem>
+                    <para>Contains the SQL Server authentication extension,
+                        <filename>guacamole-auth-jdbc-sqlserver-0.9.13-incubating.jar</filename>,
+                        along with a <filename>schema/</filename> directory contains SQL Server-specific
+                        scripts requires to set up the database.  The JAR extension file will need to be
+                        placed within the <filename>GUACAMOLE_HOME/extensions</filename> folder, while the
+                        SQL Server JDBC driver must be placed within the <filename>GUACAMOLE_HOME/lib</filename>
+                        directory.</para>
+                    <para><emphasis>The SQL Server JDBC driver is not included with the extension.</emphasis>  You
+                        must obtain the JDBC driver <filename>.jar</filename> yourselfand place it in the directory.
```
mike-jumper: yourselfand

src/chapters/jdbc-auth.xml
```diff
@@ -97,9 +97,29 @@
...
+                        Furthermore, the SQL Server authentication extension supports a number of TDS-compatible
+                        drivers, so you must make sure the one you choose is supported by the extension, that the
+                        extension is configured properly, and that the <filename>.jar</filename> is in the correct
+                        directory.  Microsoft's JDBC driver can be downloaded from <link
+                        xlink:href="https://docs.microsoft.com/en-us/sql/connect/sql-connection-libraries#anchor-20-drivers-relational-access">
+                        this page</link>.</para>
```
mike-jumper: Is there a way this could be rephrased such that the link is not, simply, "this page"? I'm of the school of thought that the text of a link should represent what it's linking to.

src/chapters/jdbc-auth.xml
```diff
@@ -299,6 +321,60 @@ Type "help" for help.
                     to run any database upgrade scripts.</para>
             </important>
         </section>
+        <section xml:id="jdbc-auth-sqlserver">
+            <title>SQL Server</title>
+            <para>If using SQL Server, the database and schema must be created first.  The example below assumes
+                that you are running SQL Server on Linux, using the command-line tools to manage it.</para>
```
mike-jumper: I suggest also mentioning that any tool which can execute SQL against the SQL Server database should suffice.

src/chapters/jdbc-auth.xml
```diff
@@ -299,6 +321,60 @@ Type "help" for help.
...
+<prompt>1></prompt> <userinput>CREATE DATABASE <replaceable>guacamole_db</replaceable>;</userinput>
+<prompt>2></prompt> <userinput>GO</userinput>
+<prompt>1></prompt> <userinput>CREATE LOGIN <replaceable>guacamole_user</replaceable> WITH PASSWORD = '<replaceable>S0me_Password</replaceable>';</userinput>
+<prompt>2></prompt> <userinput>GO</userinput>
+<prompt>1></prompt> <userinput>USE <replaceable>guacamole_db</replaceable>;</userinput>
+<prompt>2></prompt> <userinput>GO</userinput>
+<prompt>1></prompt> <userinput>CREATE USER <replaceable>guacamole_user</replaceable>;</userinput>
+<prompt>2></prompt> <userinput>GO</userinput>
+<prompt>1></prompt> <userinput>ALTER ROLE db_datawriter ADD MEMBER <replaceable>guacamole_user</replaceable>;</userinput>
+<prompt>1></prompt> <userinput>ALTER ROLE db_datareader ADD MEMBER <replaceable>guacamole_user</replaceable>;</userinput>
```
mike-jumper: I don't have access to this at the moment to test, but based on the prompt pattern above, I'm surprised that the prompt before both of the `ALTER` statements is `1`. Does it keep incrementing until `GO`? Or is it truly `1` for both statements in this case?

src/chapters/jdbc-auth.xml
```diff
@@ -299,6 +321,60 @@ Type "help" for help.
...
+        <section xml:id="jdbc-auth-sqlserver">
+            <title>SQL Server</title>
+            <para>If using SQL Server, the database and schema must be created first.  The example below assumes
+                that you are running SQL Server on Linux, using the command-line tools to manage it.</para>
+            <informalexample>
+                <screen><prompt>$</prompt> <userinput>/opt/mssql-tools/bin/sqlcmd -S localhost -U SA</userinput>
+<prompt>Password:</prompt> <userinput><replaceable>password</replaceable></userinput>
+<prompt>1></prompt> <userinput>CREATE DATABASE <replaceable>guacamole_db</replaceable>;</userinput>
+<prompt>2></prompt> <userinput>GO</userinput>
+<prompt>1></prompt> <userinput>CREATE LOGIN <replaceable>guacamole_user</replaceable> WITH PASSWORD = '<replaceable>S0me_Password</replaceable>';</userinput>
```
mike-jumper: Any reason to avoid using the same password placeholder as the MySQL and PostgreSQL docs?

src/chapters/jdbc-auth.xml
```diff
@@ -299,6 +321,60 @@ Type "help" for help.
...
+<prompt>2></prompt> <userinput>GO</userinput>
+<prompt>1></prompt> <userinput>USE <replaceable>guacamole_db</replaceable>;</userinput>
+<prompt>2></prompt> <userinput>GO</userinput>
+<prompt>1></prompt> <userinput>CREATE USER <replaceable>guacamole_user</replaceable>;</userinput>
+<prompt>2></prompt> <userinput>GO</userinput>
+<prompt>1></prompt> <userinput>ALTER ROLE db_datawriter ADD MEMBER <replaceable>guacamole_user</replaceable>;</userinput>
```
mike-jumper: How do the `db_datawriter` and `db_datareader` roles compare to explicit grants for `SELECT`, `INSERT`, `UPDATE`, and `DELETE`?

src/chapters/jdbc-auth.xml
```diff
@@ -413,12 +509,29 @@ postgresql-password: <replaceable>some_password</replaceable></programlisting>
...
                             <entry>
                                 <para>The password Guacamole should provide when authenticating with
                                     the database. This is given as "some_password" in the examples
                                     given in this chapter.</para>
                             </entry>
                         </row>
+                        <row>
+                            <entry></entry>
+                            <entry></entry>
+                            <entry><property>sqlserver-driver</property></entry>
```
mike-jumper: Since this property is unique to SQL Server, it might be best documented separately, rather than in the table of required properties common to all.

src/chapters/jdbc-auth.xml
```diff
@@ -413,12 +509,29 @@ postgresql-password: <replaceable>some_password</replaceable></programlisting>
...
+                            <entry>
+                                <para>The SQL Server driver has an additional required parameter to
+                                configure the TDS compatibility of the driver.  This driver supports
+                                four different options for this command:
+                                - <property>microsoft2005</property> for the current Microsoft driver,
```
mike-jumper: The `<property>` tag represents a property, not a possible value of a property. In this case, you probably want `<constant>`. There are also semantic tags for various types of lists here, if you want to represent this as a list, as the dashes will not result in this formatting as you might expect; they'll all be inline.

Formatting them inline is probably fine, honestly, but if that's the intent, I'd say remove the dashes.

mike-jumper:
Please also update the merge base to `staging/0.9.14`, as [GUACAMOLE-363](https://issues.apache.org/jira/browse/GUACAMOLE-363) is part of 0.9.14 scope.

necouchman:
> Please also update the merge base to staging/0.9.14, as GUACAMOLE-363 is part of 0.9.14 scope.
Done.  Will try to update this today so it can be merged, soon.

src/chapters/jdbc-auth.xml
necouchman: I guess if it's a placeholder it doesn't matter, since we want people to change it from the default placeholder.  SQL Server does not allow passwords as simple as the placeholder, which is why I had tweaked it.  But I'll put it back.

src/chapters/jdbc-auth.xml
necouchman: According to Microsoft's documentation ([1]):

> db_datawriter | Members of the db_datawriter fixed database role can add, delete, or change data in all user tables.
> db_datareader | Members of the db_datareader fixed database role can read all data from all user tables.

Lacking expert SQL Server skills, but having tried a couple of things, it does not look like there's a way to wildcard the GRANT in SQL Server the way you can in MySQL or "ALL TABLES" as in Postgres.  According a thread I found ([2]) you can grant permissions across the entire schema, so could do something like:

`GRANT SELECT,INSERT,UPDATE,DELETE ON SCHEMA::[dbo] TO guacamole_user`

but it sounds like this is roughly the equivalent of adding to the db_datareader and db_datawriter roles, except the roles grant those privileges across the entire DB (all schemas) vs. a single schema.  Let me know if you have a preference on which route to go - I think adding to the roles is more consistent with how MySQL and PostgreSQL examples are currently set up (grants to all tables in the entire DB).  Limiting to a particular schema is slightly more secure; however, the way the documentation/examples read, it shouldn't matter as its a separate DB with nothing else in it - a separate schema would only apply in the case where the user explicitly loads it into a custom schema (which would probably break the JDBC module, anyway).

[1] - https://docs.microsoft.com/en-us/sql/relational-databases/security/authentication-access/database-level-roles
[2] - https://social.msdn.microsoft.com/Forums/sqlserver/en-US/1489337c-56c9-4bb8-9875-3a75be7596be/grant-select-insert-update-delete-on-all-table?forum=transactsql

src/chapters/jdbc-auth.xml
necouchman: Fixed.

src/chapters/jdbc-auth.xml
necouchman: Fixed.

src/chapters/jdbc-auth.xml
necouchman: Fixed.

src/chapters/jdbc-auth.xml
necouchman: Fixed.

src/chapters/jdbc-auth.xml
necouchman: Fixed.

src/chapters/jdbc-auth.xml
necouchman: Moved.

src/chapters/jdbc-auth.xml
necouchman: Fixed.

src/chapters/jdbc-auth.xml
mike-jumper: Reading further to double-check that the permissions granted for `db_datawriter` and `db_datareader` are not beyond what is required, I found [a table listing the specific permission granted for each](https://technet.microsoft.com/en-us/library/ms189612(v=sql.90).aspx):

sql-server-roles

You're correct - these are entirely equivalent.

src/chapters/jdbc-auth.xml
```diff
@@ -299,6 +343,61 @@ Type "help" for help.
...
+<computeroutput>
+(1 rows affected)
+
+(3 rows affected)
+                        
+(5 rows affected)</computeroutput></screen>
+            </informalexample>
+            <para>If the operation is successful, the tables and permissions have been created successfully, and you
+                can now use the database with the Guacamole client web application.</para>
+            <important xml:id="jdbc-auth-sqlserver-upgrade">
```
mike-jumper: Since this isn't yet relevant, we should probably just omit this. We can re-add and update the `<important>` section when upgrade scripts exist.

src/chapters/jdbc-auth.xml
```diff
@@ -299,6 +343,61 @@ Type "help" for help.
...
+                <para>If you are upgrading from an older version of Guacamole and were already using
+                    SQL Server, you may need to run one or more database schema upgrade scripts
+                    located within the <filename>schema/upgrade/</filename> directory. Each of these
+                    scripts is named
+                        <filename>upgrade-pre-<replaceable>VERSION</replaceable>.sql</filename>
+                    where <replaceable>VERSION</replaceable> is the version of Guacamole where those
+                    changes were introduced. They need to be run when you are upgrading from a
+                    version of Guacamole older than <replaceable>VERSION</replaceable>.</para>
+                <para>SQL Server support was added in version <replaceable>VERSION</replaceable>, so, at the time of this
```
mike-jumper: The wording here is odd, referring to the specific version where SQL Server support was added with the placeholder that is meant to refer to whichever version is current:

added-in-version-version

If the `<important>` section is just removed overall, however, that that will address this, too.

src/chapters/jdbc-auth.xml
```diff
@@ -97,9 +97,51 @@
...
+                        directory.  Microsoft's JDBC driver can be downloaded from Microsoft's <link
+                        xlink:href="https://docs.microsoft.com/en-us/sql/connect/sql-connection-libraries#anchor-20-drivers-relational-access">
+                        SQL Connection Libraries</link> page.</para>
+                    <para>In addition to the various parameters mentioned below, the SQL Server driver has a
+                        unique parameter available to control the driver compatibility of the JDBC module:
+                        <property>sqlserver-driver</property>.  This parameter allows you to choose the compatibility
+                        mode of the module with various TDS-comptabile drivers such that it can be used with different
+                        versions of SQL Server and even non-Microsoft SQL Server databases.  The following options are available
+                        for the <property>sqlserver-driver</property> property:
+                        <variablelist>
```
mike-jumper: Not sure what's going on, but the list of values isn't actually appearing in the HTML output of the manual build:

section-contents-missing

I'm guessing that it might have to do with the `<variablelist>` being inside the `<para>`, but not sure.

necouchman: Looks like I was missing the `<listitem>` tags.  Also changed it from `<variablelist>` to `<itemizedlist>` to get the desired bullet effect.

src/chapters/jdbc-auth.xml
necouchman: Removed.

src/chapters/jdbc-auth.xml
necouchman: Removed.

src/chapters/jdbc-auth.xml
```diff
@@ -346,38 +435,53 @@ mysql-hostname: localhost
...
                 relatively few and self-explanatory, describing only how the connection to the
                 database is to be established, and how Guacamole will authenticate when querying the
                 database:</para>
             <informaltable frame="all">
-                <tgroup cols="3">
+                <tgroup cols="4">
                     <colspec colname="c1" colnum="1" colwidth="1*"/>
                     <colspec colname="c2" colnum="2" colwidth="1*"/>
-                    <colspec colname="c3" colnum="3" colwidth="2*"/>
+                    <colspec colname="c3" column="3" colwidth="1*"/>
```
mike-jumper: The attribute name here should actually be `colnum`, not `column`.

src/chapters/jdbc-auth.xml
```diff
@@ -97,9 +97,51 @@
...
+                        SQL Connection Libraries</link> page.</para>
+                    <para>In addition to the various parameters mentioned below, the SQL Server driver has a
+                        unique parameter available to control the driver compatibility of the JDBC module:
+                        <property>sqlserver-driver</property>.  This parameter allows you to choose the compatibility
+                        mode of the module with various TDS-comptabile drivers such that it can be used with different
+                        versions of SQL Server and even non-Microsoft SQL Server databases.  The following options are available
+                        for the <property>sqlserver-driver</property> property:
+                        <itemizedlist>
+                            <listitem>
+                                <constant>microsoft2005</constant>: The current Microsoft driver,
```
mike-jumper: Though this does now render, it does not validate against the DocBook schema, as neither text nor `<constant>` are allowed children (see http://tdg.docbook.org/tdg/5.0/listitem.html). Wrapping the contents with `<para>` should fix things.

src/chapters/jdbc-auth.xml
```diff
@@ -97,9 +97,51 @@
...
+                        directory.  Microsoft's JDBC driver can be downloaded from Microsoft's <link
+                        xlink:href="https://docs.microsoft.com/en-us/sql/connect/sql-connection-libraries#anchor-20-drivers-relational-access">
+                        SQL Connection Libraries</link> page.</para>
+                    <para>In addition to the various parameters mentioned below, the SQL Server driver has a
+                        unique parameter available to control the driver compatibility of the JDBC module:
+                        <property>sqlserver-driver</property>.  This parameter allows you to choose the compatibility
+                        mode of the module with various TDS-comptabile drivers such that it can be used with different
+                        versions of SQL Server and even non-Microsoft SQL Server databases.  The following options are available
+                        for the <property>sqlserver-driver</property> property:
+                        <itemizedlist>
```
mike-jumper: Why the switch to `<itemizedlist>`? Since this is a list of terms and their definitions/descriptions, `<variablelist>` seems the most semantically sound:

http://tdg.docbook.org/tdg/5.0/variablelist.html

necouchman: Okay, switched back to `<variablelist>`.

src/chapters/jdbc-auth.xml
necouchman: Okay, I switched over to `<term>` for the `<varlistentry>` items - hopefully that looks okay?

src/chapters/jdbc-auth.xml
necouchman: Fixed.

src/chapters/jdbc-auth.xml
```diff
@@ -97,9 +97,54 @@
...
+                    <para>In addition to the various parameters mentioned below, the SQL Server driver has a
+                        unique parameter available to control the driver compatibility of the JDBC module:
+                        <property>sqlserver-driver</property>.  This parameter allows you to choose the compatibility
+                        mode of the module with various TDS-comptabile drivers such that it can be used with different
+                        versions of SQL Server and even non-Microsoft SQL Server databases.  The following options are available
+                        for the <property>sqlserver-driver</property> property:
+                        <variablelist>
+                            <varlistentry>
+                                <term>microsoft2005</term>
+                                <listitem>The current Microsoft driver, supporting SQL Server 2005 and later.</listitem>
```
mike-jumper: Much better, but text is still not an allowed child of `<listitem>` according to the DocBook schema: http://tdg.docbook.org/tdg/5.0/listitem.html (see the "Children" header, towards the bottom)

You'll need to wrap the contents of each `<listitem>` in `<para>` for the document to be valid.

necouchman: Done.

PR #69, opened 2018-01-01
GUACAMOLE-423: Bump version numbers in manual to 0.9.14.

mike-jumper:


src/chapters/jdbc-auth.xml
```diff
@@ -227,20 +227,13 @@ Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.
...
-<prompt>$</prompt></screen>
-                </informalexample>
+                    version of Guacamole older than <replaceable>VERSION</replaceable>.</para>
                 <para>These scripts are incremental and, when relevant, <emphasis>must be run in
                         order</emphasis>. For example, if you are upgrading an existing database
-                    from version 0.9.10-incubating, you would need to run the
-                        <filename>upgrade-pre-0.9.11.sql</filename> script (because 0.9.10 is older
-                    than 0.9.11), followed by the <filename>upgrade-pre-0.9.13.sql</filename> script
-                    (because 0.9.10 is also older than 0.9.13):</para>
+                    from version 0.9.10-incubating to version 0.9.13-incubating, you would need to
```
necouchman: Was there a reason to leave the -incubating reference in here and not make the example for 0.9.14, without that?  I think it's good either way, just want to make sure.

mike-jumper: My only reason was to avoid having to rewrite the example.

necouchman: That's fair.  Like I said, fine with me, just wanted to make sure.

mike-jumper: (...with the rest of the rewording here to avoid having to rewrite the example for future releases, as well)

PR #74, opened 2018-01-16
GUACAMOLE-464: document enable-environment-properties property

ceharris:


necouchman:
These changes look great to me.  I'm wondering if makes sense to link to this particular parameter documentation from a place or two in the rest of the documentation - for example, since this change is primarily driven by running the Guacamole Client inside of a Docker container, does it make sense to add a mention of this somewhere on the Docker install page, with a link to that location on the configuration page?  I don't want to needlessly complicate the documentation, but the Docker page does make mention of some basic configuration stuff, so not sure if this should be added in there?

ceharris:
Reading through the Docker section of the manual, it does seem like it would make some sense to weave in a mention of and a link to the new feature. 

Your comment here reminded me that the `start.sh` script that is part of the Docker image can now be modified to take advantage of the new environment properties feature. This script was actually the inspiration for the feature.

necouchman:
Ping @ceharris - Any progress on adding Docker updates to the manual?

ceharris:
Haven't forgotten about this. Just a really busy time of year at work. Hope to get back to this in the next week or so.

necouchman:
Hey, @ceharris, gentle nudge.  This issue is slated for the 1.0.0 release, so if you're able to get back to this, soon, to finish up the docs, that would be awesome.  Thanks!

ceharris:
Do you have some idea of what you'd like to see in the Docker install page for this feature?

necouchman:
> Do you have some idea of what you'd like to see in the Docker install page for this feature?

I think my initial thought was just something along the lines of "you can also allow Guacamole to be configured using environment variables, which may be useful when running it in a Docker container.  For more info, see..."  I'm sure there's better language than that, but just something that points them in the direction of the documentation on environment variable configuration.

And, as you mentioned in a previous post, this also probably means another JIRA issue for tweaking the Guacamole Client Dockerfile and start.sh script such that, in the Docker environment, the Guacamole.properties file is set to allow environment variable configuration by default and make use of the feature.  But that's for another day....

ceharris:
At long last... I've added a cross reference in the Docker chapter.

ceharris:
Changed the merge base to `staging/1.0.0`.

PR #87, opened 2018-04-01
GUACAMOLE-38: Add chapter for Quickconnect extension.

necouchman:
Adds the chapter for the quickconnect extension.


mike-jumper:
Please update the merge base to point to `staging/1.0.0` when you can.

necouchman:
Changed to staging/1.0.0.

src/chapters/quickconnect.xml
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<chapter xml:id="quickconnect" xmlns="http://docbook.org/ns/docbook" version="5.0" xml:lang="en"
+    xmlns:xi="http://www.w3.org/2001/XInclude" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <title>Quickconnect extension</title>
```
mike-jumper: Should the title represent more the high-level goal satisfied by the extension, rather than the extension itself? I don't have a suggestion for this off the top of my head, but from the perspective of a user browsing the table of contents for help, they presumably wouldn't know that the solution to their problem is the quickconnect extension until they read the chapter in question.

This is the reasoning behind the existing chapters being named things like "Database authentication" rather than "JDBC extension", or "Proxying Guacamole" rather than "Apache and Nginx".

src/chapters/quickconnect.xml
```diff
@@ -0,0 +1,90 @@
...
+                can be used. <emphasis>Doing this will disconnect all active users, so be sure
+                    that it is safe to do so prior to attempting installation.</emphasis>
+                When ready, restart your servlet container and give the extension a try.</para>
+        </section>
+    </section>
+    <section xml:id="using-quickconnect">
+        <title>Using the quickconnect extension</title>
+        <para>The quickconnect extension provides a field on the home page that allows you to enter
+             a Uniform Resource Identifier (URI) to create a connection.  A URI is in the form:</para>
+            <informalexample><replaceable>protocol</replaceable>://<replaceable>username</replaceable>
```
mike-jumper: There should be a Docbook tag which represents the semantics of a URI which can be nested within `<informalexample>`. Currently, this is simply normal paragraph text.

src/chapters/quickconnect.xml
```diff
@@ -0,0 +1,90 @@
...
+            :<replaceable>password</replaceable>@<replaceable>host</replaceable>:
+            <replaceable>port</replaceable>/?<replaceable>options</replaceable></informalexample>
```
mike-jumper: The documentation for the supported protocols refers to these as "parameters", not "options". I suggest:

* Consistently referencing the connection parameters as parameters
* Providing a concrete example or set of examples which demonstrates the use of these parameters
* Explaining, at least briefly, how multiple parameters are represented in the URI (`name1=value1&name2=value2&...`), as that may not be clear.

mike-jumper:
As there are security implications to using this extension, we should add a warning or similar text to the beginning of the chapter which makes these aspects clear. There are frequent, unwary users that "simply want to allow users to connect anywhere" but likely do not fully understand that doing so is dangerous unless they fully trust those users:

* Allowing users to connect anywhere includes machines they may not actually expect or want users to try to connect to, such as the Guacamole server itself.
* Some connection parameters allow files to be read from or transferred to arbitrary locations (through support for file transfer and screen recording, for example). Unless the permissions of the server are properly locked down, this may expose the Guacamole server to tampering.

mike-jumper:
For example, consider:

    rdp://someReachableServer/?enable-drive=true&drive-path=/etc

If the administrator doesn't fully trust the users effectively creating arbitrary connections, and the administrator has not locked down permissions on the server such that guacd cannot access/modify such things, the above would entirely expose `/etc`.

This isn't an issue in Guacamole, but definitely something that users may not fully appreciate when considering using this extension.

src/chapters/quickconnect.xml
necouchman: Some ideas:
- Adhoc Connections
- User-defined Connections
- Temporary Connections

I'm having a hard time finding anything that just calls out to me, here...

src/chapters/quickconnect.xml
necouchman: Yep, the `<uri/>` tag.  I've added it in - should make it a little prettier.

src/chapters/quickconnect.xml
necouchman: Okay, I changed all of the appropriate references to parameters to use parameter instead of option, so that should be good.  Also provided three examples of URIs, including a RDP one with several parameters specified in the URI.

necouchman:
> As there are security implications to using this extension, we should add a warning or similar text to the beginning of the chapter which makes these aspects clear.

Your points are well-taken, and I re-worked the "Important" section at the beginning of the chapter into bullet points and added a warning for the security implications.

Do you think it warrants any more generic documentation in the Guacamole Server section on overall security - that is, what user guacd runs as and the implications of that, jailing guacd, etc.?  I'm happy to take a stab at writing some documentation on that...

necouchman:
@mike-jumper Any further changes you want to see to this?

src/chapters/adhoc-connections.xml
```diff
@@ -0,0 +1,118 @@
...
+        will parse the URI and immediately establish the connection.  The purpose of the extension is
+        to allow situations where administrators want to allow users the flexibility of establishing
+        their own connections without having to grant them access to edit connections or even to have
+        to create the connections at all, aside from typing the URI.</para>
+    <important>
+        <para>There are several implications of using this extension that should be well-understood
+            by administrators prior to implementing it:</para>
+        <itemizedlist>
+            <listitem><para>Connections established with this extension are created in-memory
+                    and only persist until the user logs out or the browser is closed.</para></listitem>
```
mike-jumper: > ... or the browser is closed.

Is this the case? I thought the connections persist in server-side memory, within the user's session, independently of whether the browser has been closed.

necouchman: Modified that a little.

PR #90, opened 2018-07-02
GUACAMOLE-524: Document added ldap-user-attributes tokens

jaredfrees:


src/chapters/configuring.xml
```diff
@@ -4033,6 +4033,18 @@ guaclog: INFO: All files interpreted successfully.</computeroutput>
...
                     </listitem>
                 </varlistentry>
+                <varlistentry>
+                    <term><varname>${GUAC_ATTR_<replaceable>CUSTOM_ATTRIBUTE_NAME</replaceable>}</varname></term>
+                    <listitem>
+                        <para>An attribute value specified on the current Guacamole user.
+                            <replaceable>CUSTOM_ATTRIBUTE_NAME</replaceable> is a custom attrbiute
+                            name specified in guacamole.properties. The values for these attributes
+                            are queried from the LDAP directory when the user is authenticated.
+                            There is a token for each attribute specified in guacamole.properties,
```
necouchman: Well, than *can* be queried from LDAP, but that's not the only place they can come from - they are simply user attributes that may be obtained from any of the authentication modules that implement that feature.  LDAP has an option for which ones to query (as you've documented in the LDAP section), but any of the user attributes from the JDBC module should also be available, and I have a PR for the CAS module to pull in attributes provided by CAS SSO.

Also, only in the case of the LDAP module will the attributes be specified in guacamole.properties - in the JDBC module they are already part of the user record, and in the CAS module the available attributes are determined by the CAS server and what is is configured to provide.

A long way of saying, probably this section here should be a little more generic :smiley:.

src/chapters/ldap-auth.xml
```diff
@@ -356,6 +356,17 @@ dn: cn={4}guacConfigGroup,cn=schema,cn=config
...
                 </varlistentry>
+                <varlistentry>
+                    <term><property>ldap-user-attributes</property></term>
+                    <listitem>
+                        <para>The attribute or attributes to retrieve from the LDAP directory.
+                            These attributes are stored as tokens with the prefix "GUAC_ATTR_"
+                            and the name of the attribute appended in uppercase letters.
+                            The value of the token is the value of the attribute in the LDAP directory.
+                            If the attribute has no value in the directory then the token is not saved.
+                            Multiple attributes can be specified here, separated by commas.</para>
```
necouchman: Two things:
- Probably should clarify here that the attribute(s) are for the currently logged-in user.
- Also probably want to address how multi-valued attributes behaved.  This occurred to me at one point during the phase of writing the code, but left my mind, but it should at least be documented.  For example, in general LDAP allows the "mail" attribute to contain multiple values, and at one of my previous employers there was a large portion of the company that had multiple e-mail addresses.  What happens in this case?  There are several LDAP attributes that fall into that same category that can have multiple values, and it should at least be documented what value gets passed into this token (first value, last value, random value, all of them together...?).

jaredfrees: If the LDAP attribute has multiple values then the token would get all of the values together. I would assume someone using this feature wouldn't use a attribute with multiple values? If they did, I'm not sure what we could change to help them pick which value to use. Do you have an idea?

necouchman: I wouldn't assume that someone would not use an attribute with multiple values - they may either not be aware or not have control of that.

> If they did, I'm not sure what we could change to help them pick which value to use.

I don't know that we'd be able to help control what of the multiple values is used, but, in the `getLDAPAttributes()` method, instead of using

    attr.getStringValue()

we could use:

    attr.getStringValueArray()[0]

Probably have to do a couple of other checks to make sure that there's actually a value there (we don't get a null array or empty value, but probably worth changing the code slightly to do that, and then just document that, if the attribute is multi-valued, it'll return the first value.

jaredfrees: So it turns out getStringValue() actually returns the the first value of the attribute as a String. You can find it here:
https://www.novell.com/documentation/developer/jldap/jldapenu/api/
So it already returns the first value. I also just tested this and it does return just the first value of the attribute.

necouchman: Sounds good.

src/chapters/ldap-auth.xml
```diff
@@ -356,6 +356,20 @@ dn: cn={4}guacConfigGroup,cn=schema,cn=config
...
+                    <listitem>
+                        <para>The attribute or attributes to retrieve from the LDAP directory
+                            for the currently logged-in user. These attributes are stored as
+                            tokens with the prefix "GUAC_ATTR_" and the name of the attribute appended
+                            in uppercase letters. The value of the token is the value of the attribute
+                            in the LDAP directory for the currently logged-in user. If the attribute
+                            has no value in the LDAP directory then the token is not saved.
+                            If the attribute has multiple values in the LDAP directory then the token
+                            saves the first value of the attribute. Multiple attributes can be
+                            specified here, separated by commas.</para>
```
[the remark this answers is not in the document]
jaredfrees: How do I link to the tokens section specifically? I see that to link to Chapter 5: Configuring Guacamole I can do this: `<xref linkend="configuring-guacamole"/>`

jaredfrees:
Of course I spelled "examples" wrong in the commit and didn't notice until I pushed, haha.

jaredfrees:
Let me know if I did the links wrong.

src/chapters/ldap-auth.xml
```diff
@@ -356,6 +356,32 @@ dn: cn={4}guacConfigGroup,cn=schema,cn=config
...
+                        <para>For example, if <property>ldap-user-attributes</property> is
+                        "<systemitem>mail, workstation</systemitem>", then a GUAC_ATTR_MAIL
```
necouchman: Formatting here should be consistent with the paragraph above- that is the `<para>` tag is good where it is, but the next lines should be indented.

src/chapters/ldap-auth.xml
```diff
@@ -356,6 +356,32 @@ dn: cn={4}guacConfigGroup,cn=schema,cn=config
...
+                        token would be set to the value of the mail attribute in the LDAP directory
+                        for the currently logged-in user and a GUAC_ATTR_WORKSTATION token
+                        would be set to the value of the workstation attribute similarly,
+                        contingent on the fact that the attributes have a value in the LDAP directory.
+                        So, the tokens could be used like this:
+                        <varname>${GUAC_ATTR_MAIL}</varname> or <varname>${GUAC_ATTR_WORKSTATION}/<varname>.
```
necouchman: Typo, here - `/<varname>` should be `</varname>`.

PR #92, opened 2018-08-16
GUACAMOLE-470: Document new color-scheme parameter format for SSH and Telnet

darchons:
Document the new format of the color-scheme parameter with some examples, for SSH and Telnet connections.

### DIFF
--- a/src/chapters/configuring.xml
+++ b/src/chapters/configuring.xml
@@ -3009,13 +3009,81 @@ ed272546-87bd-4db9-acba-e36e1a9ca20a
                                             <primary>SSH</primary>
                                             <secondary>color scheme</secondary>
                                         </indexterm>The color scheme to use for the terminal
-                                        emulator used by SSH connections. Each color scheme dictates
-                                        the default foreground and background color for the
-                                        terminal. Programs which specify colors when printing text
-                                        will override these defaults.</para>
-                                    <para>This parameter is optional. By default, Guacamole will
-                                        render text as gray over a black background.</para>
-                                    <para>Possible values are:</para>
+                                        emulator used by SSH connections. It consists of a
+                                        semicolon-separated series of name-value pairs. Each
+                                        name-value pair is separated by a colon and assigns a
+                                        value to a color in the terminal emulator palette. For
+                                        example, to use blue text on white background by default,
+                                        and change the red color to a purple shade, you would
+                                        specify:</para>
+                                    <informalexample>
+                                        <programlisting>foreground: rgb:00/00/ff;
+background: rgb:ff/ff/ff;
+color9: rgb:80/00/80</programlisting>
+                                    </informalexample>
+                                    <para>This format is similar to the color configuration format
+                                        used by Xterm, so Xterm color configurations can be easily
+                                        adapted for Guacamole. This parameter is optional. If not
+                                        specified, Guacamole will render text as gray over a black
+                                        background.</para>
+                                    <para>Possible color names are:</para>
+                                    <variablelist>
+                                        <varlistentry>
+                                            <term><constant>foreground</constant></term>
+                                            <listitem>
+                                                <para>Set the default foreground color.</para>
+                                            </listitem>
+                                        </varlistentry>
+                                        <varlistentry>
+                                            <term><constant>background</constant></term>
+                                            <listitem>
+                                                <para>Set the default background color.</para>
+                                            </listitem>
+                                        </varlistentry>
+                                        <varlistentry>
+                                            <term><constant>color&lt;n&gt;</constant></term>
+                                            <listitem>
+                                                <para>Set the color at index <code>&lt;n&gt;</code>
+                                                    on the Xterm 256-color palette. For example,
+                                                    <code>color9</code> refers to the red color.
+                                                    </para>
+                                            </listitem>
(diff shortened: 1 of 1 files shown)


src/chapters/configuring.xml
```diff
@@ -3009,13 +3009,102 @@ ed272546-87bd-4db9-acba-e36e1a9ca20a
...
+                                                <para>Use the 24-bit color specified in hexadecimal.
+                                                    For example, <code>rgb:ff/00/00</code> specifies
+                                                    the color red.</para>
+                                            </listitem>
+                                        </varlistentry>
+                                        <varlistentry>
+                                            <term><constant>rgb:RRR/GGG/BBB</constant></term>
+                                            <listitem>
+                                                <para>Use a 36-bit color specified in hexadecimal.
+                                                    The color value is truncated to 24 bits. For
```
necouchman: A couple of things, here:
- Since the value is always going to be 24-bits, either zero-padded or truncated, does it make sense to documented four separate ways to specify the color values, or does it make more sense to just document it as 24-bits, with a few extra notes (or examples) for what happens if you specify it as 12, 36, or 48 bits?  I'm genuinely asking, here - maybe it does make sense to document it this way?  @mike-jumper @jmuehlner, thoughts?
- In either case, whether leaving it split up like this or collapsing it a bit, it would probably make sense to have some idea of the practical implications of what the truncating, in particular, does.  So, if I specify a 36-bit color value, and it gets truncated to 24-bits, how is that going to impact the color I've entered?

src/chapters/configuring.xml
```diff
@@ -3009,13 +3009,102 @@ ed272546-87bd-4db9-acba-e36e1a9ca20a
...
                                         </indexterm>The color scheme to use for the terminal
-                                        emulator used by SSH connections. Each color scheme dictates
-                                        the default foreground and background color for the
-                                        terminal. Programs which specify colors when printing text
-                                        will override these defaults.</para>
-                                    <para>This parameter is optional. By default, Guacamole will
-                                        render text as gray over a black background.</para>
-                                    <para>Possible values are:</para>
+                                        emulator used by SSH connections. It consists of a
+                                        semicolon-separated series of name-value pairs, separated by
```
necouchman: I'd suggest a slight tweak to this language - perhaps something like:

> It consists of a semicolon-separated series of name-value pairs, where each name-value pair is separate by a colon, ...

or something else that clarifies this a little bit.  My initial take on this is, "Well, is it separate by a semicolon or a colon?"  I get it when I look at the examples, but I think it has the potential to cause some confusion.

src/chapters/configuring.xml
```diff
@@ -3009,13 +3009,102 @@ ed272546-87bd-4db9-acba-e36e1a9ca20a
...
+                                        a colon, that assign default values to colors in the
+                                        terminal emulator palette. Note that programs that specify
+                                        colors when printing text can still override these defaults.
```
necouchman: How does this work in practice?  I understand that if you specify `foreground` and `background` that those could be overridden by the program (e.g. `ls -l --colors=auto`, which would list a directory with some color-coding outside of the specified foreground color), but, if, as in your example below, you change `color9` to a purple shade, presumably any application that goes to use `color9` to print red would, instead, print in purple?

src/chapters/configuring.xml
```diff
@@ -3009,13 +3009,102 @@ ed272546-87bd-4db9-acba-e36e1a9ca20a
...
+                                        default, and change the default red color to a purple shade,
+                                        you would specify:</para>
+                                    <informalexample>
+                                        <programlisting>foreground: rgb:00/00/ff;
+background: rgb:ff/ff/ff;
+color9: rgb:80/00/80</programlisting>
+                                    </informalexample>
+                                    <para>This format is similar to the color configuration format
+                                        used by Xterm, so Xterm color configurations can be easily
+                                        adapted for Guacamole. This parameter is optional. If not
```
necouchman: How does that "adaption" work?  Is there anything special that needs to be done when translating from Xterm color configurations to Guacamole?  If so, might make sense to put another `<informalexample>` block in right here.

mike-jumper:
The merge base for this will need to be changed to `staging/1.0.0` as GUACAMOLE-470 is part of 1.0.0.

src/chapters/configuring.xml
darchons: Collapsed to one section, with examples for zero-extension and truncation.

src/chapters/configuring.xml
darchons: Fixed

src/chapters/configuring.xml
darchons: Yeah that was language left over from the current manual. Re-reading it, I don't think it adds much value to the section, so I decided to delete that sentence altogether.

src/chapters/configuring.xml
darchons: It goes from something like

    xterm*color9: rgb:cf/61/71
    xterm*color10: rgb:7c/bc/8c

to

    color9: rgb:cf/61/71;
    color10: rgb:7c/bc/8c;

I only mentioned it as a hint that users can look for premade color schemes from xterm configs. I think the conversion process is straightforward and obvious enough that it's unnecessary to include more details/examples.

PR #93, opened 2018-08-16
GUACAMOLE-565: Document terminal-type parameter for SSH and Telnet

darchons:
Document the new terminal-type parameter for SSH and Telnet connections.

### DIFF
--- a/src/chapters/configuring.xml
+++ b/src/chapters/configuring.xml
@@ -3142,6 +3142,18 @@ ed272546-87bd-4db9-acba-e36e1a9ca20a
                                         sends ASCII code 127 (Delete) if this option is not set.</para>
                                 </entry>
                             </row>
+                            <row>
+                                <entry><parameter>terminal-type</parameter></entry>
+                                <entry>
+                                    <para><indexterm>
+                                            <primary>SSH</primary>
+                                            <secondary>terminal type</secondary>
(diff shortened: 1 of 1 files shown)


src/chapters/configuring.xml
```diff
@@ -3142,6 +3142,24 @@ ed272546-87bd-4db9-acba-e36e1a9ca20a
...
                                 </entry>
                             </row>
+                            <row>
+                                <entry><parameter>terminal-type</parameter></entry>
+                                <entry>
+                                    <para><indexterm>
+                                            <primary>SSH</primary>
+                                            <secondary>terminal-type</secondary>
+                                        </indexterm>This parameter sets the terminal emulator type
+                                        string that is passed to the SSH server. Possible values
```
necouchman: I'm not sure it's worth listing out all of the possible values, here.  Since this is being documented for the Guacamole Client interface the values are in a drop-down box and the user will be able to see them.  It should be sufficient just to document the purpose and what happens if you don't specify it, and leave the list off.

mike-jumper:
The merge base for this will need to be changed to `staging/1.0.0` as GUACAMOLE-565 is part of 1.0.0.

src/chapters/configuring.xml
darchons: Fixed

PR #99, opened 2018-10-28
GUACAMOLE-220: Update database documentation for user groups.

mike-jumper:
These changes:

* Reorganize the database auth documentation to include SQL Server among the databases affected by database upgrades (as the user group changes will be the first schema changes for the SQL Server support)
* Update the schema documentation to include the new tables, changes to the columns of existing tables, etc. added/made as a result of the user group changes.

To make things easier to navigate, the reorganization includes separating the documentation into distinct sections for creating the initial database, upgrading an existing database, and granting access to the database for Guacamole, with database-specific subsections for these, rather than main database-specific sections with repeated create/upgrade/grant subsections.

### DIFF
--- a/src/chapters/jdbc-auth.xml
+++ b/src/chapters/jdbc-auth.xml
@@ -737,15 +737,72 @@ sqlserver-user-required: true</programlisting>
         <para>This section assumes knowledge of SQL and your chosen database, and that whatever you
             need to do can be accomplished if only you had high-level information about Guacamole's
             SQL schema.</para>
+        <section xml:id="jdbc-auth-schema-entities">
+            <title>Entities</title>
+            <indexterm>
+                <primary><classname>guacamole_entity</classname></primary>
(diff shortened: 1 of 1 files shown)


necouchman:
This is not part of the section that has been changed, but since you're already brushing this up, there are a couple of typos in the password complexity section:

https://github.com/apache/guacamole-manual/blob/cc2da2ed3b391ecafe55a6f7fab7bd92e379393c/src/chapters/jdbc-auth.xml#L548

Maybe those could be fixed as part of these changes, too?

necouchman:
Also, wording seems a little off, here:

https://github.com/apache/guacamole-manual/blob/cc2da2ed3b391ecafe55a6f7fab7bd92e379393c/src/chapters/jdbc-auth.xml#L594-L595

mike-jumper:
Sure. I'll fix and reword a bit.

PR #105, opened 2019-01-23
Added HTML lang="en"

skotfred:
Primarily an accessibility concern for screen reader software.

### DIFF
--- a/tutorials/guacamole-tutorial/src/main/webapp/index.html
+++ b/tutorials/guacamole-tutorial/src/main/webapp/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE HTML>
-<html>
+<html lang="en">
 
     <head>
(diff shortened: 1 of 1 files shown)


necouchman:
Sorry for the delay on this @skotfred.  A couple of things:
- All pull requests need a matching JIRA issue, so you'll need to put in a JIRA issue for this.
- Is this the only place this needs to be done?  Are there any other accessibility things we should consider concurrent to this change?  If not, that's fine, just figure while we're doing it we might as well look around...

mike-jumper:
I think this change is not in the intended place - it affects only the `index.html` within the bundled example source of the webapp tutorial, not the HTML generated when the manual itself is built.

With respect to the generated HTML, this should be addressed by #165. Sphinx automatically adds the appropriate `lang` attribute to its output.

PR #108, opened 2019-02-02
GUACAMOLE-718: Correct slight grammatical error in TOTP chapter.

necouchman:
Because we only should speak like that on Speak Like a Pirate Day.


necouchman:
@mike-jumper @ceharris @jmuehlner @JoelB

Anyone care to review and merge?!  This one is easy, I promise!

PR #109, opened 2019-02-09
GUACAMOLE-729: Bump version numbers to 1.1.0.

mike-jumper:


mike-jumper:
Definitely wrong. I'll fix that.

PR #111, opened 2019-06-21
GUACAMOLE-422: Document support for timezone handshake and parameter.

necouchman:
A few items covered by this pull request:
* Document the timezone parameters for SSH and RDP
* Rearrange SSH documentation to match how parameters are currently configured.
* Document the timezone portion of the handshake, along with some notes about how instructions can now be in arbitrary order during the handshake phase.

I'm sure it needs some work, particularly the protocol/handshake section.

### DIFF
--- a/src/chapters/configuring.xml
+++ b/src/chapters/configuring.xml
@@ -2969,6 +2969,23 @@ ed272546-87bd-4db9-acba-e36e1a9ca20a
                                     <para>For more information, please see <xref linkend="ssh-host-verification"/>.</para>
                                 </entry>
                             </row>
+                            <row>
+                                <entry><parameter>server-alive-interval</parameter></entry>
+                                <entry>
+                                    <para>
+                                        <indexterm>
+                                            <primary>SSH</primary>
+                                            <secondary>server-alive-interval</secondary>
+                                        </indexterm>
+                                        By default the SSH client does not send keepalive requests
+                                        to the server.  This parameter allows you to configure the
+                                        the interval in seconds at which the client connection
+                                        sends keepalive packets to the server.  The default is 0,
+                                        which disables sending the packets.  The minimum value
+                                        is 2.
+                                    </para>
+                                </entry>
+                            </row>
                         </tbody>
                     </tgroup>
                 </informaltable>
@@ -3224,9 +3241,13 @@ color9: rgb:80/00/80</programlisting>
                 </informaltable>
             </section>
             <section xml:id="ssh-command">
-                <title>Session parameters</title>
-                <para>The parameters in this section control various aspects
-                of the SSH session.</para>
+                <title>Running a command (instead of a shell)</title>
+                <para>By default, SSH sessions will start an interactive shell. The shell which will
+                    be used is determined by the SSH server, normally by reading the user's default
+                    shell previously set with <command>chsh</command> or within
+                        <filename>/etc/passwd</filename>. If you wish to override this and instead
+                    run a specific command, you can do so by specifying that command in the
+                    configuration of the Guacamole SSH connection.</para>
                 <informaltable frame="all">
                     <indexterm>
                         <primary>parameters</primary>
@@ -3248,71 +3269,9 @@ color9: rgb:80/00/80</programlisting>
                                     <para><indexterm>
                                             <primary>SSH</primary>
                                             <secondary>command</secondary>
-                                        </indexterm>By default, SSH sessions will
-                                        start an interactive shell.  The shell
-                                        that is used will be determined by the
-                                        SSH server, normally by reading the user's
-                                        default login shell specified in
-                                        <filename>/etc/passwd</filename>.  If you
-                                        wish to override this and run a specific
-                                        command, you can do so by filling in a
-                                        value for this parameter.  Assuming the
-                                        server allows arbitrary commands to be
-                                        run, the session will attempt to start
-                                        the specified command in place of the
-                                        shell.  This parameter is optional. If not specified, the SSH
+                                        </indexterm>The command to execute over the SSH session, if
+                                        any. This parameter is optional. If not specified, the SSH
                                         session will use the user's default shell.</para>
-                                </entry>
-                            </row>
-                            <row>
-                                <entry><parameter>timezone</parameter></entry>
-                                <entry>
-                                    <para>
-                                        <indexterm>
-                                            <primary>SSH</primary>
-                                            <secondary>timezone</secondary>
-                                        </indexterm>This parameter allows you
-                                        to control the timezone that is sent
-                                        to the server over the SSH connection,
-                                        which will change the way local time is
-                                        displayed on the server.</para>
-                                    <para>
-                                        The mechanism used to do this over SSH
-                                        connections is by setting the
-                                        <envar>TZ</envar> variable on the SSH
-                                        connection to the timezone specified by
-                                        this parameter.  This means that the SSH
-                                        server must allow the <envar>TZ</envar>
-                                        variable to be set/overriden - many SSH
-                                        server implementations have this disabled
-                                        by default.  To get this to work, you may
-                                        need to modify the configuration of the
-                                        SSH server and explicitly allow for
-                                        <envar>TZ</envar> to be set/overriden.
-                                    </para>
-                                    <para>
-                                        The available values of this parameter are
-                                        standard IANA key zone format timezones,
-                                        and the value will be sent directly to
-                                        the server in this format.
-                                    </para>
-                                </entry>
-                            </row>
-                            <row>
-                                <entry><parameter>server-alive-interval</parameter></entry>
-                                <entry>
-                                    <para>
-                                        <indexterm>
-                                            <primary>SSH</primary>
-                                            <secondary>server-alive-interval</secondary>
-                                        </indexterm>
-                                        By default the SSH client does not send keepalive requests
-                                        to the server.  This parameter allows you to configure the
-                                        the interval in seconds at which the client connection
-                                        sends keepalive packets to the server.  The default is 0,
-                                        which disables sending the packets.  The minimum value
-                                        is 2.
-                                    </para>
                                 </entry>
                             </row>
                         </tbody>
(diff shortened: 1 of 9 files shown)


src/chapters/protocol.xml
```diff
@@ -80,6 +80,22 @@
             it supports Ogg Vorbis audio, but no video, and can accept both PNG and JPEG images. It
             wants to connect to localhost at port 5900, and is leaving the three other parameters
             blank.</para>
+        <para>Version 1.1.0 of Guacamole adds a couple of other features to the handshake phase.  First,
```
mike-jumper: I'm not sure about listing new changes within the documentation as a set of layered changes. I feel that the documentation for version X really should be a reference for version X first, with other relevant information (protocol is different in older versions) being secondary.

What do you think of instead describing things strictly from the perspective of the way things are in this version? I think it would make more sense for the section to describe the protocol of the latest version with a note about the behavior of older versions, rather than a section mainly describing older versions with a note about the behavior of the latest.

src/chapters/protocol.xml
```diff
@@ -80,6 +80,22 @@
...
+            that the client has finished sending options.  In addition to ordering, this means
+            that certain instructions can be omitted without preventing the connection from
+            continuing.
+        </para>
+        <para>Also introduced in version 1.1.0 is the ability to pass the client timezone to
+            guacd, which can be used by protocols that support it to alter how the server
+            displays local time for the logged in user.  More information on this is available
+            in <xref linkend="configuring"/>, under the documentation related to
+            the timezone parameter.  If either the client or server do not support the
+            timezone instruction in the handshake the information will not be pased and the
```
mike-jumper: The new `timezone` instruction will also need to be documented in the protocol reference: [src/references/protocol.xml](https://github.com/apache/guacamole-manual/blob/6df18f34fbb156120725176dfa914c6f59d79751/src/references/protocol.xml).

src/chapters/protocol.xml
```diff
@@ -80,6 +80,22 @@
...
             blank.</para>
+        <para>Version 1.1.0 of Guacamole adds a couple of other features to the handshake phase.  First,
+            the handshake can now be done with most of the handshake instructions sent in arbitrary order.
+            Beginning with this version, guacd will detect and parse out each of the options regardless
+            of the order they are sent in, looking for the connect instruction as the indication
+            that the client has finished sending options.  In addition to ordering, this means
+            that certain instructions can be omitted without preventing the connection from
+            continuing.
+        </para>
+        <para>Also introduced in version 1.1.0 is the ability to pass the client timezone to
```
mike-jumper: Now that the handshake is evolving, it may be easier to just list out the handshake instructions as a table. I don't really have a problem with prose, and I generally like how you've phrased things here, but I do think things should be described in a more absolute way, not layers of relative additions.

For example, these changes currently describe things more-or-less like:

* The protocol handshake requires you to provide `size`, `audio`, `video`, and `image`.
* But actually, never mind. That's the old protocol. Those are optional now and may occur in any order.
* Wait! There's also a new `timezone` instruction you can use.

I'd lean more towards:

* The protocol handshake consists of any of the following instructions, which are each optional and may occur in any order: (instructions....)
* **NOTE:** Older releases supported a smaller subset of these instructions and required handshake instructions to be in a specific order. See the archived versions of the manual for the protocol documentation relevant to older releases.

mike-jumper:
Is the protocol version part of the handshake intended to be a part of this?

necouchman:
> Is the protocol version part of the handshake intended to be a part of this?

I can go either way, but I figured since it was merged in under the same pull request for the server and client portions, it was fine to include it in the manual section.

src/chapters/configuring.xml
```diff
@@ -1741,6 +1741,48 @@ tcp6       0      0 :::4713                 :::*                    LISTEN</comp
...
+                                    timezone varies by RDP server implementation.
+                                    For example, with Windows, support for
+                                    forwarding timezones is only present in
+                                    Windows Server with Remote Desktop Services
+                                    (RDS, formerly known as Terminal Services)
+                                    installed.  Windows Server installations in
+                                    admin mode, along with Windows workstation
+                                    versions, do not allow the timezone to be
+                                    forwarded.  Other server implementations,
+                                    for example, xRDP, may not implement this
```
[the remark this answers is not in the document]
necouchman: Or maybe all lower-case?  The official website seems to use lower-case...

http://www.xrdp.org/

src/chapters/protocol.xml
necouchman: Yeah, that's a good point.  I'll re-word it to just describe the `1.1.0` behavior.

src/chapters/protocol.xml
necouchman: Okay.

src/chapters/protocol.xml
necouchman: Sounds good - will try some different language.

mike-jumper:
> I can go either way, but I figured since it was merged in under the same pull request for the server and client portions, it was fine to include it in the manual section.

Definitely, but if that's the intent, I don't see it here (`VERSION_1_1_0`, etc.).

necouchman:
> Definitely, but if that's the intent, I don't see it here (VERSION_1_1_0, etc.).

Ah, yep, you're right.  Will try to add that in with the other changes.

src/chapters/protocol.xml
necouchman: Added.

src/chapters/configuring.xml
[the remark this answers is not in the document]
necouchman: Went with "xrdp"

src/chapters/protocol.xml
necouchman: Should be good, now.

src/chapters/protocol.xml
necouchman: Okay, have taken a stab at shamelessly copying a table from elsewhere in the manual and substituting in the handshake instructions.  Let me know how it looks - currently compiling it to make sure everything formats well, but I believe I at least got valid XML.

src/chapters/protocol.xml
```diff
@@ -57,29 +57,90 @@
...
+            using an "args" instruction:</para>
         <informalexample>
-            <programlisting>4.args,8.hostname,4.port,8.password,13.swap-red-blue,9.read-only;</programlisting>
+            <programlisting>4.args,13.VERSION_1_1_0,8.hostname,4.port,8.password,13.swap-red-blue,9.read-only;</programlisting>
         </informalexample>
+        <para>The protocol version is used to negotiate compatibility between differing
+            versions of client and server, allowing the two sides to negotiate the highest
+            supported version and enable or disable features associated with that version.
+            Older versions of the Guacamole Client that do not support this instruction
+            will silently ignore it as an empty connection parameter.  Valid protocol versions
```
mike-jumper: Valid protocol versions (and their general format) should probably be documented with the instruction in the protocol reference. A good example of this might be the `ack` instruction, which also documents all legal status codes:

https://github.com/apache/guacamole-manual/blob/33e653ccf99cbb7b978719b6ec2cc07a0e8bed30/src/references/instructions/server/stream/ack.xml#L32-L225

mike-jumper:
Beware that the merge base for this will also need to be `staging/1.1.0`, as GUACAMOLE-422 is part of the 1.1.0 scope.

src/references/instructions/server/handshake/args.xml
```diff
@@ -9,9 +9,16 @@
...
         the handshake phase only.</para>
-    <para>The parameters of the args instruction are the names of all parameters
-        accepted by the server for the protocol in selected by the client, in
-        order. The client's responding connect instruction must the values of
-        each of these parameters in the same order.</para>
+    <para>The first parameter of this instruction will be the protocol version
+        supported by the server.  This is used to negotiate protocol
+        compatibility between the client and the server, with the highest
+        supported protocol by both sides being chosen.  Versions of Guacamole
+        prior to 1.1.0 do not support this instruction and cannot
```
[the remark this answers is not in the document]
necouchman: Should be more accurate, now.

src/references/instructions/client/handshake/timezone.xml
```diff
@@ -0,0 +1,14 @@
...
+
+<section xml:id="timezone-handshake-instruction"
+    xmlns="http://docbook.org/ns/docbook" version="5.0" xml:lang="en"
+    xmlns:xi="http://www.w3.org/2001/XInclude">
+    <title>timezone</title>
+    <indexterm>
+        <primary>timezone</primary>
+    </indexterm>
+    <para>Specifies the timezone of the client system, in IANA zone key format.
+        This is a single-value parameter, and may be used by protocols to
```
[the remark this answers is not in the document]
necouchman: Added.

src/chapters/protocol.xml
necouchman: Added a table with the versions.

necouchman:
Okay, rebased to `staging/1.1.0` and corrected those issues.

src/chapters/protocol.xml
```diff
@@ -57,29 +57,125 @@
...
+                            <entry>Version</entry>
+                            <entry>Value</entry>
+                            <entry>Description</entry>
+                        </row>
+                    </thead>
+                    <tbody>
+                        <row>
+                            <entry>1.0.0</entry>
+                            <entry><constant>VERSION_1_0_0</constant></entry>
+                            <entry>This is the defualt version and applies to any
```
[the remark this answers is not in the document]
necouchman: a before u expect after...

necouchman:
Removed that extra content and verified that it compiles this time.

src/chapters/protocol.xml
```diff
@@ -57,29 +57,122 @@
...
-            and the values for all arguments available, even if blank. If any of these requirements
-            are left out, the connection will close:</para>
+            and the values for all arguments available, even if blank.</para>
         <informalexample>
             <programlisting>4.size,4.1024,3.768,2.96;
 5.audio,9.audio/ogg;
 5.video;
 5.image,9.image/png,10.image/jpeg;
+8.timezone,16.America/New_York;
 7.connect,9.localhost,4.5900,0.,0.,0.;</programlisting>
```
mike-jumper: Is this still correct with the addition of `VERSION_1_1_0` to the corresponding `args`?

src/chapters/configuring.xml
```diff
@@ -3201,13 +3224,9 @@ color9: rgb:80/00/80</programlisting>
                 </informaltable>
             </section>
             <section xml:id="ssh-command">
```
mike-jumper: If this section is no longer specific to running a command, the ID should be updated accordingly, as it's this ID that will be used to link to specific sections.

src/chapters/configuring.xml
```diff
@@ -3229,11 +3248,73 @@ color9: rgb:80/00/80</programlisting>
...
+                                    <para>
+                                        The available values of this parameter are
+                                        standard IANA key zone format timezones,
+                                        and the value will be sent directly to
+                                        the server in this format.
+                                    </para>
+                                </entry>
+                            </row>
+                            <row>
+                                <entry><parameter>server-alive-interval</parameter></entry>
```
mike-jumper: Why within session parameters vs. network parameters?

src/chapters/configuring.xml
```diff
@@ -3229,11 +3248,73 @@ color9: rgb:80/00/80</programlisting>
...
+                                        value for this parameter.  Assuming the
+                                        server allows arbitrary commands to be
+                                        run, the session will attempt to start
+                                        the specified command in place of the
+                                        shell.  This parameter is optional. If not specified, the SSH
                                         session will use the user's default shell.</para>
                                 </entry>
                             </row>
+                            <row>
+                                <entry><parameter>timezone</parameter></entry>
```
mike-jumper: If unsure about repurposing the "running a command" section, this could be documented within the "locale" section added recently via #117.

src/chapters/protocol.xml
necouchman: Nope, should contain the version in that string...

src/chapters/configuring.xml
necouchman: Yeah, I have no clue what I was doing, here.  My only guess is that I didn't have enough :coffee: before writing this...

src/chapters/configuring.xml
necouchman: :coffee: ??

src/chapters/configuring.xml
necouchman: Yes, this makes complete sense.

PR #120, opened 2019-08-12
GUACAMOLE-579: Add documentation for CAS extension tokens.

necouchman:
Initial cut at some documentation for extension-specific tokens.

### DIFF
--- a/src/chapters/configuring.xml
+++ b/src/chapters/configuring.xml
@@ -5181,6 +5181,9 @@ guaclog: INFO: All files interpreted successfully.</computeroutput>
         </section>
         <section xml:id="parameter-tokens">
             <title>Parameter tokens</title>
+            <indexterm>
+                <primary>tokens</primary>
+            </indexterm>
             <para>The values of connection parameters can contain "tokens" which will be replaced by
                 Guacamole when used. These tokens allow the values of connection parameters to vary
                 dynamically by the user using the connection, and provide a simple means of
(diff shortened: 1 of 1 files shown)


src/chapters/configuring.xml
```diff
@@ -5254,6 +5257,57 @@ guaclog: INFO: All files interpreted successfully.</computeroutput>
...
+                        tokens.  The CAS server must be specifically configured to release certain
+                        attributes to the client (Guacamole), and configuration of that is outside
+                        the scope of this document.  Any attribute that the CAS server is
+                        configured to release should be available to Guacamole as a token for
+                        use within a connection.  The token name will be prepended with the
+                        <constant>CAS_</constant> prefix.  A CAS server configured to release
+                        attributes <varname>firstname</varname>, <varname>lastname</varname>,
+                        <varname>email</varname>, and <varname>mobile</varname> would produce the
+                        following tokens:</para>
+                    <variablelist>
```
mike-jumper: The [DocBook `<variablelist>` element](https://tdg.docbook.org/tdg/4.5/variablelist.html) is defined as:

> A list in which each entry is composed of a set of one or more terms and an associated description

If there is no associated definition for each `<term>`, then a different type of list would be better, such as [`<itemizedlist>`](https://tdg.docbook.org/tdg/4.5/itemizedlist.html).

necouchman: Okay, swapped over to `<itemizedlist>`.

PR #122, opened 2019-08-13
GUACAMOLE-524: Document LDAP parameter tokens with other parameter tokens

necouchman:
This adds some parameter token documentation for the LDAP extension to the same Parameter Tokens section.

### DIFF
--- a/src/chapters/configuring.xml
+++ b/src/chapters/configuring.xml
@@ -5307,6 +5307,52 @@ guaclog: INFO: All files interpreted successfully.</computeroutput>
                         </listitem>
                     </itemizedlist>
                 </section>
+                <section xml:id="ldap-tokens">
+                    <title>LDAP Extension Tokens</title>
+                    <indexterm>
+                        <primary>tokens</primary>
+                        <secondary>ldap</secondary>
+                    </indexterm>
+                    <para>The LDAP extension will read user attributes provided by the LDAP
+                        server and specified in the <filename>guacamole.properties</filename>
+                        file.  The attributes retrieved for a user are configured using the
+                        <parameter>ldap-user-attributes</parameter> parameter.  The user must
+                        be able to read the attribute values from their own LDAP object.
+                        The token name will be prepended with the <constant>LDAP_</constant>
+                        prefix.  As an example, configuring the following line in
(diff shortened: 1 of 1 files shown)


src/chapters/configuring.xml
```diff
@@ -5307,6 +5307,50 @@ guaclog: INFO: All files interpreted successfully.</computeroutput>
...
+                    <programlisting>ldap-user-attributes: cn, givenName, sn, mobile, mail</programlisting>
+                    <itemizedlist>
+                        <listitem>
+                            <para>
+                                <varname>${LDAP_CN}</varname>
+                            </para>
+                        </listitem>
+                        <listitem>
+                            <para>
+                                <varname>${CAS_GIVENNAME}</varname>
```
mike-jumper: Copypasta - this and a few others have a `CAS_` prefix.

src/chapters/configuring.xml
```diff
@@ -5307,6 +5307,50 @@ guaclog: INFO: All files interpreted successfully.</computeroutput>
...
+                    <para>The LDAP extension will read user attributes provided by the LDAP
+                        server and specified in the <filename>guacamole.properties</filename>
+                        file.  The attributes retrieved for a user are configured using the
+                        <param>ldap-user-attributes</param> parameter.  The user must be able
+                        to read the attribute values from their own LDAP object.  The token
+                        name will be prepended with the <constant>LDAP_</constant> prefix.  As
+                        an example, configuring the following line in
+                        <filename>guacamole.properties</filename> file will produce the
+                        below tokens that can be used in connection configurations:</para>
+                    <programlisting>ldap-user-attributes: cn, givenName, sn, mobile, mail</programlisting>
```
mike-jumper: I think `<programlisting>` normally needs to be within `<informalexample>` when used in this way. I would also recommend splitting things across where they are actually referenced, as it feels a bit odd to read a sentence which notes and is followed by two examples.

src/chapters/configuring.xml
necouchman: Oops, fixed.

src/chapters/configuring.xml
necouchman: Okay, added `<informalexample>` and split the paragraph.

PR #125, opened 2020-01-06
GUACAMOLE-249: Update documentation with respect to FreeRDP 2.x support.

mike-jumper:
This change updates the manual with respect to the FreeRDP 2.x support added through apache/guacamole-server#243. In particular:

* Notes warning that a particular feature is unavailable in an older version of FreeRDP have been removed (as only FreeRDP 2.x or above is now supported).
* Build output, build requirements, and error messages have been updated to reflect changes in log messages made via apache/guacamole-server#243.
* The new "nla-ext" security mode has been added, and the documentation of existing security modes has been clarified or expanded.
* The default security mode has been updated from "rdp" to "any".


jmuehlner:
Looks good to me.

PR #127, opened 2020-02-24
GUACAMOLE-300: documentation for new property 'ldap-member-attribute-type'

mlewissmith:


src/chapters/ldap-auth.xml
```diff
@@ -369,6 +369,15 @@ dn: cn={4}guacConfigGroup,cn=schema,cn=config
                           here.</para>
                     </listitem>
                 </varlistentry>
+                <varlistentry>
+                    <term><property>ldap-member-attribute-type</property></term>
+                    <listitem>
+                          <para>Specify whether the attribute defined in
```
necouchman: It feels like this is written in "short-hand" rather than following the style of the rest of the documentation.  Maybe a little more verbosity ;-).

necouchman:
@mlewissmith: Ping - any chance you can re-work this documentation a bit so we can get it merged?

necouchman:
@mike-jumper: You good with this documentation as written?

mike-jumper:
Sure, I'm OK with moving forward with the latest revision.

PR #133, opened 2020-05-26
GUACAMOLE-1081: Document token case modifiers.

necouchman:
A stab at documentation for the new token modifiers - I'm sure this will need a little brushing up, so let me know what needs to be changed.

### DIFF
--- a/src/chapters/configuring.xml
+++ b/src/chapters/configuring.xml
@@ -6001,13 +6001,14 @@ guaclog: INFO: All files interpreted successfully.</computeroutput>
                 forwarding authentication information without storing that information in the
                 connection configuration itself, so long as the remote desktop connection uses the
                 same credentials as Guacamole.</para>
-            <para>Each token is of the form
-                    <varname>${<replaceable>TOKEN_NAME</replaceable>}</varname>, where
-                    <replaceable>TOKEN_NAME</replaceable> is some descriptive name for the value the
-                token represents. Tokens with no corresponding value will never be replaced, but
-                should you need such text within your connection parameters, and wish to guarantee
-                that this text will not be replaced with a token value, you can escape the token by
-                adding an additional leading "$", as in "$${TOKEN_NAME}".</para>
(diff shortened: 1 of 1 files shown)


mike-jumper (commented):
Is it still planned that these modifiers will change to `UPPER` and `LOWER` respectively?

necouchman:
> Is it still planned that these modifiers will change to UPPER and LOWER respectively?

Yeah - just lost track of it.  Have changed the documentation and will submit a new PR on the client-side for this...

src/chapters/configuring.xml
```diff
@@ -6002,12 +6002,13 @@ guaclog: INFO: All files interpreted successfully.</computeroutput>
                 connection configuration itself, so long as the remote desktop connection uses the
                 same credentials as Guacamole.</para>
             <para>Each token is of the form
-                    <varname>${<replaceable>TOKEN_NAME</replaceable>}</varname>, where
-                    <replaceable>TOKEN_NAME</replaceable> is some descriptive name for the value the
-                token represents. Tokens with no corresponding value will never be replaced, but
-                should you need such text within your connection parameters, and wish to guarantee
-                that this text will not be replaced with a token value, you can escape the token by
-                adding an additional leading "$", as in "$${TOKEN_NAME}".</para>
+                    <varname>${<replaceable>TOKEN_NAME</replaceable>:<replaceable>MODIFIER</replaceable>}</varname>,
```
mike-jumper, line 6005: I think this should be clarified to document the form as "`${TOKEN_NAME}` or `${TOKEN_NAME:MODIFIER}`", to avoid giving the impression that a modifier is required.

mike-jumper (changes requested):

src/chapters/configuring.xml
necouchman: Yeah, good point. I've modified it to include some language that should clarify that.

mike-jumper (approved):

PR #144, opened 2020-06-25
GUACAMOLE-728: Update JDBC SSL mode documentation

necouchman:


src/chapters/jdbc-auth.xml
```diff
@@ -100,7 +100,13 @@
                     </para>
                     <para>The MySQL Driver implements several parameters specific to configuring
                         SSL for secure connections to MySQL servers that support or require
-                        encrypted communications.  The parameters are as follows:</para>
+                        encrypted communications. The SSL configuration options are targeted
+                        at newer versions of the MySQL Connector/J and MariaDB Connector/J.
+                        Older versios of MySQL Connector/J have known issues with SSL
```
mike-jumper, line 105: versions*

For the sake of readability, I would additionally:

* Wrap that main "older versions have known issues" bit within `<emphasis>`, as it's fairly important
* Clean up the repeated "Connector/J" a bit, as it's hard to read with all those slashes and repetition.

 For example:

> The MySQL Driver implements several parameters specific to configuring SSL for secure connections to MySQL servers that support or require encrypted communications. _Older versions of Connector/J driver have known issues with SSL verification_. If you experience problems connecting to SSL-secured MySQL or MariaDB databases, it is recommended that you update to a current version of the driver.
>
> Configuration parameters for MySQL-compatible SSL support are as follows:

necouchman:
I believe I did get the Postgres default update in there - down toward the bottom?

mike-jumper:
Ah, you're right. I just needed to click the arrows a bit.

PR #149, opened 2020-11-02
GUACAMOLE-221: Document connection parameter prompting.

necouchman:
Seems like there isn't all that much to update for parameter prompting, so hopefully this takes care of it. Let me know if anything else needs to be added/adjusted.

### DIFF
--- a/src/chapters/configuring.xml
+++ b/src/chapters/configuring.xml
@@ -1635,15 +1635,18 @@ tcp6       0      0 :::4713                 :::*                    LISTEN</comp
                     domain parameters are omitted. One notable exception to this is Network Level
                     Authentication, or NLA, which performs all authentication outside of a desktop
                     session, and thus in the absence of a graphical interface.</para>
-                <important>
-                    <para>If your server requires NLA, you <emphasis>must</emphasis> provide a
-                        username and password. Leveraging Guacamole's <link
-                            xmlns:xlink="http://www.w3.org/1999/xlink" linkend="parameter-tokens"
-                            >parameter tokens</link> and <link
-                            xmlns:xlink="http://www.w3.org/1999/xlink" linkend="ldap-auth">LDAP
-                            support</link> to integrate with Active Directory and automatically pass
-                        through credentials is a common configuration.</para>
-                </important>
+                <para>Servers that require NLA can be handled by Guacamole in one of two ways. The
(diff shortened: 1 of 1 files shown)


src/chapters/configuring.xml
```diff
@@ -6195,6 +6195,33 @@ guaclog: INFO: All files interpreted successfully.</computeroutput>
...
+                authenticators were not provided as part of the connection configuration.
+            </para>
+            <important>
+                <para>It is important to note that requests for parameters will only be
+                    generated in the case where that information has not already been
+                    provided as part of the connection. The user will never be asked for
+                    parameters that replace or override connection parameters where
+                    values have been configured as part of the connection, including
+                    authentication information. For example, if a connection to a
+                    RDP server specifies a username and password, and that username
```
mike-jumper, line 6219: I think this should be clarified to refer to a "connection configuration" or "the configuration of a connection" or similar, rather than just "connection".

src/chapters/configuring.xml
```diff
@@ -6195,6 +6195,33 @@ guaclog: INFO: All files interpreted successfully.</computeroutput>
...
+                    or password is incorrect and results in an  authentication failure,
+                    the connection will fail and the user will not be prompted for
+                    alternative credentials.</para>
```
mike-jumper, line 6222: > ... the user will not be prompted for alternative credentials.

Some RDP servers will still take it upon themselves to display a login screen and prompt for correct credentials (if NLA is not in use). It's good to assure the admin that Guacamole will not prompt the user for these things, but I think we should be more specific with that assurance.

mike-jumper (changes requested):

mike-jumper (commented):
Looks good.

Thoughts on other parts of the docs referencing NLA? I think the following is no longer correct, and there may be others:

> If your server requires NLA, you must provide a username and password. Leveraging Guacamole's parameter tokens and LDAP support to integrate with Active Directory and automatically pass through credentials is a common configuration.

necouchman:
> Thoughts on other parts of the docs referencing NLA? I think the following is no longer correct

Sure, I'll take a look through and make sure those get cleaned up.

necouchman:
Okay, I've updated a couple of the other sections that talk about NLA to include some words on prompting.

mike-jumper (approved):

PR #151, opened 2020-11-21
GUACAMOLE-1219: Allow TOTP to be disabled by group membership

janhf:
Adds the documentation for the new configuration parameter.


necouchman:
Please note that this now needs to be re-worked in the new Sphinx format of the manual.

PR #160, opened 2021-02-22
GUACAMOLE-747: Add verbiage to point users to example branding code.

necouchman:


src/chapters/guacamole-ext.xml
```diff
@@ -308,6 +308,10 @@
                 its contents will automatically be inserted directly below the login dialog (the
                 only element that would match <code>.login-ui .login-dialog</code>) as if they were
                 part of Guacamole's HTML in the first place.</para>
+            <para>An example of an extension that modifies style and HTML components for the purpose
+                of providing custom "branding" of the Guacamole Client interface can be found in the
```
mike-jumper, line 312: I think this should be "Guacamole interface" rather than "Guacamole Client interface".

src/chapters/guacamole-ext.xml
```diff
@@ -308,6 +308,10 @@
...
+                <filename>doc/guacamole-branding-example</filename> directory of the Guacamole
+                Client source code.</para>
```
mike-jumper, line 314: For clarity, I think this should be "the guacamole-client source code". Might also be helpful to link to: https://github.com/apache/guacamole-client/tree/master/doc/guacamole-branding-example

mike-jumper (changes requested):

src/chapters/guacamole-ext.xml
necouchman: Hopefully the way I did it makes sense - let me know if the wording is too awkward.

mike-jumper (approved):

PR #174, opened 2021-12-12
GUACAMOLE-1364: Document SSO extension structure and behavior.

mike-jumper:
This change:

* Documents the new SSO extension structure adopted as of 1.4.0.
* Describes how SSO extensions behave when they have priority vs. when they don't, as well as how to change that priority to achieve the desired behavior.

### DIFF
--- a/src/saml-auth.md
+++ b/src/saml-auth.md
@@ -106,6 +106,31 @@ the scope of this document, and will vary widely based on the IdP in use.
   management within Guacamole Client, particularly when layered with other
   authentication modules. This property is optional, and defaults to "groups".
 
+### Controlling login behavior
+
+```{include} include/sso-login-behavior.md
+```
+
+#### Automatically redirecting all unauthenticated users
+
+To ensure users are redirected to the SAML identity provider immediately
+(without a Guacamole login screen), ensure the SAML extension has priority over
+all others:
+
+```
+extension-priority: saml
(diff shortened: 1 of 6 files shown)


src/saml-auth.md
```diff
@@ -15,15 +15,12 @@ management.
...
-versions of Guacamole here: <http://guacamole.apache.org/releases/>.
+```{include} include/sso-download.md
+```
 
-The SAML authentication extension is packaged as a `.tar.gz` file containing
-only the extension itself, `guacamole-auth-saml-1.4.0.jar`, which must
-ultimately be placed in `GUACAMOLE_HOME/extensions`.
+The extension for the desired SSO method, in this case
+`guacamole-auth-sso-openid-1.4.0.jar` from within the `openid/` subdirectory,
+must ultimately be placed in `GUACAMOLE_HOME/extensions`.
```
necouchman, line 23: Looks like a little copy-pasta, here - this should be "saml" and not "openid".

necouchman (changes requested):

src/saml-auth.md
mike-jumper: Oopsies

mike-jumper: OK - fixed via rebase.

necouchman (approved):

PR #176, opened 2021-12-29
GUACAMOLE-724: Document usage of connection tiling and keyboard focus.

mike-jumper:


src/using-guacamole.md
```diff
@@ -82,20 +82,24 @@ again or swipe left across the screen.
 
 The Guacamole menu provides options for:
 
-* Navigating back to the home screen
+* [Navigating back to the home screen](client-user-menu)
```
jmuehlner, line 85: Have you considered putting these shortcuts in the same order as the sections below?

mike-jumper: That does sound sensible.

mike-jumper: There.

PR #180, opened 2022-01-14
GUACAMOLE-1501: Makefile: use `find` in a POSIX-compatible way.

dstndstn:
Instead of using -path, pass paths as regular arguments.  This is POSIX-compatible.

Without this, the `find` call fails on Mac OSX 10.15:

find: illegal option -- p
usage: find [-H | -L | -P] [-EXdsx] [-f path] path ... [expression]
       find [-H | -L | -P] [-EXdsx] -f path [path ...] [expression]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,16 +33,16 @@ clean:
 	$(RM) -R build/
 
 # All files which the build depends on
-MD_FILES=$(shell find -path "./src/*" -name "*.md")
-RST_FILES=$(shell find -path "./src/*" -name "*.rst")
-PNG_FILES=$(shell find -path "./src/*" -name "*.png")
+MD_FILES=$(shell find "./src" -name "*.md")
+RST_FILES=$(shell find "./src" -name "*.rst")
+PNG_FILES=$(shell find "./src" -name "*.png")
 
 #
 # HTML manual build
 #
 
 html: build/html/index.html
-	
+
 build/html/index.html: $(PNG_FILES) $(RST_FILES) $(MD_FILES)
 	sphinx-build -b html -d build/doctrees src/ build/html
 


Makefile
```diff
@@ -33,16 +33,16 @@ clean:
 	$(RM) -R build/
 
 # All files which the build depends on
-MD_FILES=$(shell find -path "./src/*" -name "*.md")
-RST_FILES=$(shell find -path "./src/*" -name "*.rst")
-PNG_FILES=$(shell find -path "./src/*" -name "*.png")
+MD_FILES=$(shell find ./src -name "*.md")
+RST_FILES=$(shell find ./src -name "*.rst")
+PNG_FILES=$(shell find ./src -name "*.png")
```
necouchman, line 38: Two concerns I have, here:
1. Might still be good to quote the paths.
2. Are there going to be any side effects to omitting the wildcard on the `src/*` path?

necouchman (changes requested):

Makefile
dstndstn: 1. Is totally gratuitous since there are no variable expansions, wildcards, or anything going on, but whatever floats your boat (I committed this change with an incorrectly formatted commit message, then tried to fix it and made a mess of the history, oops)
2. Nope, I don't think so.  `find -path "./src/*"` says "search in `.` for paths matching `./src/*`" while `find "./src/"` says "search in `./src/`", so the latter would avoid even searching in other directories in `.` other than `src`.

dstndstn:
You can do squash merges as a Github option:
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests

necouchman:
@dstndstn And you can also squash them from the command line, in your branch ;-).

Actually, there's a reason we don't use the Merge & Squash option - we actually want a separate "Merge" commit, which that would do away with - by not using that option, we keep attribution (you get credit for your commit) and there's still a merge commit by one of the committers (me).

dstndstn:
Uh, sorry, I don't know that magical incantation.  If I do a `git rebase -i master` locally, I end up in a state where
```
Switched to branch 'guacamole-1501'
Your branch and 'origin/guacamole-1501' have diverged,
and have 1 and 4 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)
```

I was under the impression that once commits were pushed to github, making any changes to the history was a very bad idea.

necouchman:
After you do that, do "git push -f" to force push to github. Yes, changing history isn't a great idea, but, in this case, since it's just in your local branch and we haven't committed anything to the Apache side, yet, it's okay.

dstndstn:
Aha, thanks for that, that's useful to know.

necouchman (approved):

PR #181, opened 2022-01-14
GUACAMOLE-1499: Mention some caveats with VNC

dstndstn:
- mention that file transfer is not supported by VNC, so sftp is required.
- this may be too detailed -- mention that the Ubuntu 18.04 shipped package does not support copy/paste; point to an alternative.

### DIFF
--- a/src/configuring-guacamole.md
+++ b/src/configuring-guacamole.md
@@ -435,6 +435,10 @@ VNC support for Guacamole is provided by the libguac-client-vnc library, which
 will be installed as part of guacamole-server if the required dependencies are
 present during the build.
 
+VNC by itself does not provide file transfer, so you must use sftp (as
+described in [](common-sftp)) if you want to provide file upload or
+download in Guacamole.
+
 :::{note}
 In addition to the VNC-specific parameters below, Guacamole's VNC support also
 accepts the parameters of several features that Guacamole provides for multiple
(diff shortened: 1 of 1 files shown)


src/configuring-guacamole.md
```diff
@@ -750,6 +754,12 @@ only be accessed via VNC, one of these is likely your best choice. Both
 optimize window movement and (depending on the application) scrolling, giving a
 very responsive user experience.
 
+On Ubuntu Linux version 18.04, the package
+`tigervnc-standalone-server` does not support clipboard (copy/paste)
+functionality.  The `tigervncserver` packages provided by the TigerVNC
+project (https://sourceforge.net/projects/tigervnc/files/stable/) do
+support clipboard operations.
+
```
necouchman, line 762: My thought, here, is that, instead of trying to cover what the capabilities of specific distributions, versions, and VNC servers support, it might be better go more generic. Something more like `Your distribution and VNC server may not support clipboard functionality - please check the documentation or locate a VNC server/version that does.`

I'm just not sure the Guacamole Manual is the place to try to document (and keep up-to-date) the capabilities of various distributions and VNC servers. I don't think that'll scale well :-).

necouchman (changes requested):

src/configuring-guacamole.md
dstndstn: Yes, I hear you.  But there is not other documentation to check, and how is one supposed to "find one that does".  I don't know what causes a VNC server to have clipboard functionality that works or doesn't with guacamole, but it's certainly not something they document in a clear way!  Would a "known good" / "known missing some functionality" list be more palatable?  It would at least be less wordy.

necouchman: I think I'd be okay with a couple of lists like that - these VNC servers are known to work, these are not, etc.

PR #183, opened 2022-08-10
GUACAMOLE-1662: Document Dependency Package Error

wojiushixiaobai:


necouchman:
Overall looks okay to me, but maybe makes sense to put a note in there that this changes at a certain Ubuntu (and maybe Debian) version?

necouchman:
Also, you have two commits - based on the simplicity of these changes you probably want to squash them into a single commit.

wojiushixiaobai:
OK.

PR #192, opened 2023-01-23
GUACAMOLE-1293: Document protocol changes associated with sending messages to the client.

necouchman:
This PR documents the protocol changes that were made when implementing the notification of connection join and leave events, notably the changes to the client -> server handshake, and the msg instruction.

### DIFF
--- a/src/guacamole-protocol.md
+++ b/src/guacamole-protocol.md
@@ -87,6 +87,12 @@ parameter.
 
 Valid protocol versions are as follows:
 
+`VERSION_1_5_0`
+: Protocol version 1.5.0 introduced two new instructions - the `msg`
+  instruction, which is used to send arbitrary messages to the client, and
+  the `name` handshake instruction, which allows the client to set the
+  human-readable name of the user joining a connection.
+
 `VERSION_1_3_0`
 : Protocol version 1.3.0 introduced the `require` instruction, used by the
   server to indicate that the client must provide additional arguments (such
@@ -138,6 +144,9 @@ The following are valid instructions during the handshake:
 `image`
 : The image formats that the client supports, in order of preference. The
   client in the example above is supporting both PNG and JPEG.
+
+`name`
+: The human-readable name of the user joining the connection.
 
 `timezone`
 : The timezone of the client, in IANA zone key format. More information on
--- a/src/protocol-reference.rst
+++ b/src/protocol-reference.rst
@@ -1230,3 +1230,24 @@ human-readable way.
     The client is already using too many resources. Existing resources must be
     freed before further requests are allowed.
 
+Message Codes
+-------------
+
+The `msg` instruction must have at least one numeric argument that the client
+then uses to interpret the message and determine what action, if any, it
+should take based on the message. The following numeric codes are currently
+implemented for this instruction.
+
+1 (``USER_JOINED``)
+    Notifies the owner of a connection that another user has joined the
+    connection. This message is expected to include two additional
+    arguments: the guacd-generated UUID of the user who is joining the
+    connection, and the arbitrary name of the user provided by the
+    client during the handshake.
+
+2 (``USER_LEFT``)
+    Notifies the owner of a connection that another user has left the
+    connection. This message is expected to include two additional
+    arguments: the guacd-generated UUID of the user who is leaving the
+    connection, and the arbitrary name of the user provided by the
+    client during the handshake.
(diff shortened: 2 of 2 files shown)


src/protocol-reference.rst
```diff
@@ -713,6 +713,23 @@ Streaming instructions
     :arg string filename:
         The name of the file, as it would be saved on a filesystem.
 
+.. guac:instruction:: msg
+    :sent-by: server
+    :phase: interactive
+
+    Sends a message from the server (gaucd) to the client. The nature of these
```
mike-jumper, line 720: guacd*

src/protocol-reference.rst
```diff
@@ -713,6 +713,23 @@ Streaming instructions
...
+    messages is intentionally ambiguous and flexible - the message must include
+    a numeric code that the client understands and can act on, and may also
+    include one or more arguments that can be used by the client in association
+    with the message.
+
+    :arg integer msg:
+        A numeric value indicating the message that is being passed to the client.
```
mike-jumper, line 727: We should document those values, as well, similar to: https://guacamole.apache.org/doc/gug/protocol-reference.html#status-codes

src/protocol-reference.rst
```diff
@@ -713,6 +713,23 @@ Streaming instructions
...
+
+    :arg string args:
+        One or more arguments associated with the message that is being sent
+        to the client.
```
mike-jumper, line 731: Does the `msg` instruction also allows zero arguments? If so, I think this should be rephrased to "any number of arguments".

src/guacamole-protocol.md
```diff
@@ -87,6 +87,12 @@ parameter.
 
 Valid protocol versions are as follows:
 
+`VERSION_1_5_0`
+: Protocol version 1.5.0 introduced two new instructions - the "msg"
+  instruction, which is used to send arbitrary messages to the client, and
+  the "name" handshake instruction, which allows the client to set the
```
mike-jumper, line 93: `msg` and `name` here should be within backticks so that they are represented as inline code.

src/protocol-reference.rst
```diff
@@ -713,6 +713,23 @@ Streaming instructions
     :arg string filename:
         The name of the file, as it would be saved on a filesystem.
 
+.. guac:instruction:: msg
+    :sent-by: server
+    :phase: interactive
+
+    Sends a message from the server (gaucd) to the client. The nature of these
+    messages is intentionally ambiguous and flexible - the message must include
```
mike-jumper, line 721: I don't think "ambiguous" is quite right - each `msg` unambiguously represents a specific server-side event. Perhaps "broad"?

mike-jumper (changes requested):

src/protocol-reference.rst
necouchman: Fixed via rebase.

src/guacamole-protocol.md
necouchman: Fixed via rebase.

src/protocol-reference.rst
necouchman: Yeah, good point - fixed via rebase.

src/protocol-reference.rst
necouchman: Fixed via rebase.

src/protocol-reference.rst
necouchman: Okay, will add them after the status codes.

necouchman: Added a commit with the message code documentation.

src/protocol-reference.rst
```diff
@@ -1203,3 +1230,22 @@ human-readable way.
...
+    Notifies the owner of a connection that another user has joined the
+    connection. This message is expected to include one additional
+    argument, which is a string with the name of the user who has
+    joined the connection.
+
+2 (``USER_LEFT``)
+    Notifies the owner of a connection that another user has left the
+    connection. This message is expected to include one additional
+    argument, a string with the name of the user who has left the
+    connection.
```
mike-jumper, line 1251: I think these message types are missing an argument. For example, from `client.c`:

```c
/* Send user joined notification to owner. */
const char* args[] = { (const char*)joiner->user_id, (const char*)send_joiner, NULL };
return (void*) ((intptr_t) guac_protocol_send_msg(user->socket, GUAC_MESSAGE_USER_JOINED, args));
```

There are two arguments sent with the join/leave messages: a unique ID for the relevant user (assigned by guacd) and that user's arbitrary name (received by guacd during the connection handshake).

mike-jumper (changes requested):

src/protocol-reference.rst
necouchman: Yep, you are correct - fixed via rebase.

mike-jumper (approved):

PR #194, opened 2023-02-08
GUACAMOLE-462: Document installation and usage of the recording storage extension.

mike-jumper:


mike-jumper:
After this, just one more doc to go for 1.5.0...

src/recording-playback.md
```diff
@@ -0,0 +1,247 @@
...
+* It is very easy to locate and playback the recording for a session when doing
+  so only involves clicking a button in the connection history.
+
+This chapter of the documentation covers installing and using the extension
+that allows recordings stored on disk to be played back in the browser.
+
+:::{important}
+This chapter involves modifying the contents of `GUACAMOLE_HOME` - the
+Guacamole configuration directory. If you are unsure where `GUACAMOLE_HOME` is
+located on your system, please consult [](configuring-guacamole) before
```
jmuehlner, line 23: Don't you need some text in the link here? `[](configuring-guacamole)` looks like it won't produce anything on render (and doesn't when [viewing the file in github](https://github.com/apache/guacamole-manual/blob/aaa57ab69c9f14abebff560961377ba36caf35ac/src/recording-playback.md).

src/recording-playback.md
```diff
@@ -0,0 +1,247 @@
...
+such that it can be played back and reviewed later. Graphical recordings can be
+converted to video [using the `guacenc` tool](graphical-recording) (part of
+[guacamole-server](building-guacamole-server)) or can be played back directly
+in the browser in their native format using Guacamole itself. This has several
+benefits:
+
+* Recordings can be played back while the session is underway.
+* Recordings need not be re-encoded as traditional video, an
+  intensive process that often results in a larger file.
+* It is very easy to locate and playback the recording for a session when doing
```
jmuehlner, line 14: Minor nitpick: I think this should be "play back" when it's a verb. "playback" is only a noun as far as I can tell.

src/recording-playback.md
mike-jumper: Normally, yes, but not with Sphinx+Myst. This syntax results in the Myst parser automatically pulling in the title of the linked document/section.

For example:

* Source: https://github.com/apache/guacamole-manual/blob/1.4.0/src/duo-auth.md?plain=1#L10-L15
* Result: https://guacamole.apache.org/doc/1.4.0/gug/duo-auth.html

src/recording-playback.md
mike-jumper: Ah, yep - good catch.

mike-jumper: Fixed via rebase.

src/recording-playback.md
jmuehlner: Ah yes I was wondering if it was something like that.

PR #208, opened 2023-05-14
GUACAMOLE-1253: Document configuration of database batch size.

mike-jumper:


mike-jumper:
@necouchman Sure. I don't think a force push to `staging/1.5.2` to correct a commit message will cause trouble.

PR #217, opened 2023-11-07
GUACAMOLE-1878: Add docker related TOTP autentication to documentation

bogdantomasciuc:
Add documentation about the required and optional environment required when using the TOTP extension with docker similarly to how other authentication extensions are documented.


necouchman:
Thanks, @bogdantomasciuc - appreciate your effort on this. I have two requests:
* Thank you for including the Guacamole Jira issue in the title and commit message; however, it needs to be prefixed in the correct format, `GUACAMOLE-1878:`, rather than `GUACAMOLE-1878 - `. Please correct this.
* Could you also re-base this against the `staging/1.5.4` branch so that we can include it in this upcoming bugfix release?

src/guacamole-docker.md
```diff
@@ -791,6 +791,62 @@ SAML authentication extension.
...
+apps. Setting these value to something other than the defaults might make the codes
+unusable if your authenticator app does not support setting these parameters.
+:::
+
+Depending on your server's timezone, because the extension relies on working with
+short lived codes, in order to make the extension work reliable one may have to
+provide the correct timezone to the docker containers. Writing correct codes on the
+authentication page but getting authentication error might indicate a timezone issus.
+The correct timezone can be passed with the variable `TZ` as in `TZ="Europe/Bucharest"`,
+for example.
```
mike-jumper, line 848: @bogdantomasciuc, I don't believe this is correct. Copying from my comment in JIRA:

> ... the time comparisons involved in TOTP do not involve your local timezone. All calculations are done with timestamps pulled from a system clock that is _always_ UTC.

For example, here's how the codes are calculated:

1. `generate()` is invoked on `TOTPGenerator` that has been initialized with the user's TOTP key.

   https://github.com/apache/guacamole-client/blob/eb2a068f4dbc24602c906b14782bbc1c8dc52aec/extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/auth/totp/user/UserVerificationService.java#L344-L345

2. `TOTPGenerator` pulls the current system timestamp using `System.currentTimeMillis()`, which returns a timestamp based on UTC.

   https://github.com/apache/guacamole-client/blob/eb2a068f4dbc24602c906b14782bbc1c8dc52aec/extensions/guacamole-auth-totp/src/main/java/org/apache/guacamole/totp/TOTPGenerator.java#L427-L429

From [the documentation for `System.currentTimeMillis()`](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#currentTimeMillis--), that value is (emphasis added):

> the difference, measured in milliseconds, between the current time and **midnight, January 1, 1970 UTC.**

Local timezone settings have no bearing on this. Setting `TZ` will affect how command-line tools present the current system time to you, but this should have no impact at all on the calculations involved in producing a TOTP code.

bogdantomasciuc: Your comment and the provided documentation makes sense, but still, unless TZ is provided to the containers, the totp authentication fails consistently on my installation. I am using bare simple docker run commands (if it matters) with mysql backend. This is the (obviously edited :D) command:
```bash
docker run --name guacd --restart=always -v /GuacamoleRecordings:/var/lib/guacamole/recordings -e GUACD_LOG_LEVEL="info" -e TZ=Europe/Bucharest -d guacamole/guacd:latest

docker run --name guac-mysql --mount source=guac-mysql,target=/var/lib/mysql -v /tmp/scripts:/tmp/scripts -e MYSQL_ROOT_PASSWORD='SomeS3cr3tPassw0rd' -e TZ=Europe/Bucharest --restart=always -d mysql:latest

docker run --name guacamole --link guacd:guacd --link guac-mysql:mysql -v /var/log/guacamole:/home/guacamole/logs -v /GuacamoleRecordings:/var/lib/guacamole/recordings  -e MYSQL_DATABASE='guacamole-db' -e MYSQL_USER='guacamole-user' -e MYSQL_PASSWORD='MyguacamoleV3ryS3cr3tPassw0rd!ssst!' -e REMOTE_IP_VALVE_ENABLED='true' --mount source=guac-config,target=/config -e GUACAMOLE_HOME=/config -e RECORDING_SEARCH_PATH='/var/lib/guacamole/recordings' -e LOG_LEVEL='info' -e TOTP_ENABLED='true' -e TZ='Europe/Bucharest' --restart=always -d -p 8080:8080 guacamole/guacamole:latest
```

bogdantomasciuc: I have removed the paragraph relating to the timezone issue I observed in order to resolve this conversation.

bogdantomasciuc:
I messed up the branch on my end. I will close this pull, fork and make another clean pull. Will also remove the paragraph about the TZ. Might raise a separate issue about it if in a fresh install it behaves the same.

PR #218, opened 2023-11-09
GUACAMOLE-1878: Add docker related TOTP autentication to documentation

bogdantomasciuc:
Add documentation about the required and optional environment required when using the TOTP extension with docker similarly to how other authentication extensions are documented.

### DIFF
--- a/src/guacamole-docker.md
+++ b/src/guacamole-docker.md
@@ -786,6 +786,55 @@ SAML authentication extension.
   that may be helpful in debugging problems with SAML authentication. This
   is optional and defaults to false - debugging will not be enabled.
 
+(guacamole-docker-totp-auth)=
+
+### TOTP Authentication
+
+TOTP authentication can be configured to allow the Guacamole Client instance
+running in a Docker container to use a second layer of authentication using a
+two factor authenticator application and short one-time codes. More details 
+on TOTP authentication with Guacamole can be found on the [](totp-auth) page.
+
+(guacamole-totp-auth-required-vars)=
+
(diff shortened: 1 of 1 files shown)


src/guacamole-docker.md
```diff
@@ -786,6 +786,55 @@ SAML authentication extension.
...
+
+`TOTP_MODE`
+: The hash algorithm that should be used to generate TOTP codes. Legal values are
+  "sha1", "sha256", and "sha512". By default, "sha1" is used.
+
+:::{important}
+The duration and/or hash algorithm are not settable in some widely used autenticator
+apps. Setting these value to something other than the defaults might make the codes
+unusable if your authenticator app does not support setting these parameters.
+:::
```
necouchman, line 836: Two things about this block:
* I'd suggest moving this up above the list of these variables, so that it is more likely to be noticed by people reading through this section.
* `Setting these value to something`: `value` should be plural -> `values`

bogdantomasciuc: I corrected the mistake (typo) and moved the notice above.

necouchman (changes requested):
The changes look good - one final request - could you either 1) squash the two commits together, or 2) change the wording on the second commit so that it reflects the actual changes made in the commit (moving `important` block from bottom to above the list)?

bogdantomasciuc:
I never squashed before. Hopefully it's ok as I followed a tutorial to squash the commits.

bogdantomasciuc:
I am going to close pull request and make a separate clean pull again.

PR #232, opened 2024-06-15
GUACAMOLE-1290: Add documentation for SSH public key parameter.

necouchman:


src/configuring-guacamole.md
```diff
@@ -1676,6 +1676,12 @@ by prompting the user directly.
   parameter is not provided, the user will be prompted for the passphrase
   upon connecting.
 
+`public-key`
+: If SSH is using certificate-based authentication, this field allows
+  you to provide the Base64-encoded CA certificate that will be used
+  to validate the certificate of the user. This parameter is optional,
+  and by default will be null.
```
mike-jumper, line 1683: > This parameter is optional, and by default will be null.

This might be internally true, but "null" will have no meaning to the administrator. I think this should be rephrased from the perspective of an administrator whose only exposure to this parameter is the user-visible value.

necouchman: Fixed via rebase.

mike-jumper (approved):

PR #243, opened 2024-08-26
GUACAMOLE-1643: Document KSM user, connection-group level configuration.

jmuehlner:


src/vault.md
```diff
@@ -170,6 +174,28 @@ All other configuration properties are optional.
...
+In addition to the required global `ksm-config` configuration blob, Guacamole
+can also be configured with user or connection group KSM configuration, which
+will pull additional secrets _only_ when not already available. If a secret can
+be pulled using the `ksm-config` global KSM config, it will always be used.
+Failing that, if a secret is available using the connection grop config, that
+value will be used. Only when neither the global or containing connection group
+KSM configs define a secret will the user KSM config be used. Note also that
+user KSM configs will be disabled unless ``ksm-allow-user-config` is set to true.
+
+These KSM config values can be set directly in the webapp, on the connection
```
jmuehlner, line 194: Should I include screenshots of the edit pages for these? I could see them being useful, but it'd also be pretty cluttered if we included screenshots like this all the time. The screenshots would look something like the following:
image
image

mike-jumper: I think things are clear without screenshots, but we should probably add links to the documentation covering the "Preferences" screen and the connection group editor.

necouchman:
@jmuehlner Just FYI - both your commit message and your pull request have `GUACAMOLe` instead of `GUACAMOLE` (lower e).

jmuehlner:
> @jmuehlner Just FYI - both your commit message and your pull request have `GUACAMOLe` instead of `GUACAMOLE` (lower e).

Oops! Fixed.

src/vault.md
jmuehlner: Good call - linked.

jmuehlner:
Note: this PR also includes documentation updates for https://issues.apache.org/jira/browse/GUACAMOLE-1656 (per-user vault configuration). It probably doesn't make sense to try to seperate these changes, since they're so interdependent with the other KSM changes - should I update the commit message to also include `GUACAMOLE-1656`?

mike-jumper:
> Note: this PR also includes documentation updates for GUACAMOLE-1656 and GUACAMOLE-1629. It probably doesn't make sense to try to seperate these changes, since they're so interdependent with the other KSM changes - should I update the commit message to also include `GUACAMOLE-1656`?

Sure - I think it'd be good to add a note to the relevant commit(s).

jmuehlner:
> > Note: this PR also includes documentation updates for GUACAMOLE-1656 and GUACAMOLE-1629. It probably doesn't make sense to try to seperate these changes, since they're so interdependent with the other KSM changes - should I update the commit message to also include `GUACAMOLE-1656`?
> 
> Sure - I think it'd be good to add a note to the relevant commit(s).

Does this look reasonable?

mike-jumper:
Looks reasonable to me.

src/vault.md
```diff
@@ -115,6 +115,10 @@ question should be able to access.
 
    ![Application creation confirmation dialog showing the generated one-time token.](images/vault-ksm-004-generate-token.png)
 
+   This token can be used to generate a base-encoded configuration blob as
```
mike-jumper, line 118: base64*-encoded?

jmuehlner: Good catch - fixed.

mike-jumper (approved):

PR #249, opened 2024-08-27
GUACAMOLE-1290: Document SFTP public key parameter.

necouchman:


src/configuring-guacamole.md
```diff
@@ -2185,6 +2185,12 @@ transfer features.
   authentication. This parameter is not needed if the private key does not
   require a passphrase.
 
+`sftp-public-key`
+: If the SFTP connection is using key-based authentication, this field
+  allows you to specify the Base64-encoded CA certificate against which
```
jmuehlner, line 2190: CA certificate? Is that right? That doesn't look like what's happening here https://github.com/apache/guacamole-server/pull/538

necouchman: Yeah, that's not quite right. Looking at the documentation for the `libssh2_userauth_publickey_frommemory()` function, it looks like:
* You can pass just a private key, in which case libssh2 will extract (compute) the public key based on the private key contents.
* You can pass the private key and a separate public key, in which case the separate public key takes precedence over extraction from the private key.

If the public key data doesn't match the private key, the authentication will fail, so it's a way to pass those separately such that you're validating the private key that is passed against a specific, known public key value.

So, I need to reword this - and the other instance of it - a little differently.

necouchman: Okay, let me know if it makes a bit more sense, now.

jmuehlner: Looks reasonable to me.

PR #250, opened 2024-08-28
GUACAMOLE-1026: Update documentation to match revised FreeRDP-related output from configure.

mike-jumper:


src/installing-guacamole.md
```diff
@@ -435,6 +435,7 @@ guacamole-server version 1.5.5
       guacenc .... yes
       guaclog .... yes
 
+   FreeRDP plugins: /usr/lib/pkgconfig/../../lib/freerdp3
```
jmuehlner, line 438: It's a little weird to show freerdp2 in the output above with freerdp3 paths here, no?

mike-jumper: Oops, yes. I'll make this consistent.

mike-jumper: Fixed via rebase.

jmuehlner (approved):

PR #258, opened 2024-09-09
GUACAMOLE-1020: Document authentication restriction extension.

necouchman:


src/auth-restrict.md
```diff
@@ -0,0 +1,302 @@
...
+balancing connection groups may or not may not be used. The goal is to
+give administrators of a Guacamole system additional flexibility in
+restricting when and from where the system can be used.
+
+As this extension decorates underlying extensions, it must be used alongside
+one that is capable of storing additional information for users, user groups,
+connections, and connection groups. Currently the only extension provided
+by the Guacamole project that is capable of doing this is the
+[JDBC authentication extension](jdbc-auth). Other extensions, such as the
+[LDAP authentication extension](ldap-auth) or
```
jmuehlner, line 22: Are you saying that the LDAP and JSON extensions _are_ currently capable of being used with this extension (or might some day be?) If it's that they _might_ someday work, I'd just leave that bit out. I don't see a lot of benefit to speculatation on what features might theoretically be added to these in the future. If they would currently work, I'd be more clear about that.

src/auth-restrict.md
```diff
@@ -0,0 +1,302 @@
...
+Guacamole.
+
+#### Restricting logins based on day and time
+
+This extension introduces the ability to restrict logins to the Guacamole
+system based on the day of the week and the time of day, and to combine these
+restrictions in a way that allows for flexibility in those schedules.
+
+As an example, let's say that you have a user or group of users whose access
+you'd like to restrict to your normal working hours of 9 AM to 5 PM, any day
```
jmuehlner, line 126: It might be worth being clear about what timezone these restrictions are in - UTC? The timezone of the admin editing the page? The configured server time? The user's timezone?

necouchman: Ah, yes, you're absolutely right - I knew I was forgetting something.

src/auth-restrict.md
necouchman: Fair point. I've removed it.

src/auth-restrict.md
necouchman: Okay, added a blurb on that - let me know how it looks. Timezones and Daylight Savings Time get to be so much fun...

jmuehlner: Nice, looks good.

jmuehlner:
Maybe should have waited for https://github.com/apache/guacamole-client/pull/830 to be merged first...

If that somehow doesn't make it into the release we'll have to revert these changes.

PR #260, opened 2024-10-03
GUACAMOLE-1239: Add documentation for configuring username case sensitivity.

necouchman:
This documents case-sensitivity settings for everything except the SSL SSO module, which is still pending documentation...

### DIFF
--- a/src/cas-auth.md
+++ b/src/cas-auth.md
@@ -57,7 +57,7 @@ their identity has been confirmed:
 : The URI to redirect back to upon successful authentication. Normally this
   will be the full URL of your Guacamole installation.
 
-Additional optional properties are available to control how CAS tokens are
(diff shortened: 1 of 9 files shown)


mike-jumper:
> everything except the SSL SSO module, which is still pending documentation...

I can probably just include documentation for the new property with that overall documentation for SSL SSO.

mike-jumper (changes requested):
The following properties look to be missing:

* `json-case-sensitive-usernames`
* `radius-case-sensitive-usernames`

necouchman:
> The following properties look to be missing:
> 
> * `json-case-sensitive-usernames`
> * `radius-case-sensitive-usernames`

Added those, also created a link point for the global option so that the property can be easily accessed from each of the extension pages.

mike-jumper (approved):

PR #262, opened 2024-10-17
GUACAMOLE-1993: Provide built manual as github artifact as part of PR build.

jmuehlner:


mike-jumper (approved):
LGTM - I'm going to close and reopen to trigger this and check out the resulting artifact.

mike-jumper:
That works quite nicely.

.github/workflows/pr-build.yml
```diff
@@ -26,4 +26,16 @@ jobs:
       - name: Build Docker container
         shell: sh
         run: |
-          docker build --pull --no-cache --force-rm .
+          docker build --pull --no-cache --tag built-manual-image .
+          docker create --name built-manual-container built-manual-image
+          docker cp built-manual-container:/usr/local/apache2/htdocs "$RUNNER_TEMP/htdocs"
+          docker container rm --force --volumes built-manual-container
+          docker image rm --force built-manual-image
```
mike-jumper, line 33: Just occurred to me that this will be a race condition if multiple PR builds end up getting assigned to the same runner. Maybe the container name and image should be unique for the PR?

mike-jumper (changes requested):

.github/workflows/pr-build.yml
```diff
@@ -26,4 +26,16 @@ jobs:
       - name: Build Docker container
         shell: sh
         run: |
-          docker build --pull --no-cache --force-rm .
+          docker build --pull --no-cache --tag built-manual-image .
+          docker create --name built-manual-container built-manual-image
+          docker cp built-manual-container:/usr/local/apache2/htdocs "$RUNNER_TEMP/htdocs"
+          docker container rm --force --volumes built-manual-container
```
mike-jumper, line 32: I think you can avoid keeping the temporary container across multiple commands by replacing this with something like:

```
docker run --rm built-manual-image tar -C /usr/local/apache2 -c htdocs | tar -C "$RUNNER_TEMP" -x
```

PR #267, opened 2025-02-25
GUACAMOLE-2037: Make listen address configurable

pv2b:
Documentation updates related to https://github.com/apache/guacamole-server/pull/578


mike-jumper:
Similar to apache/guacamole-server#578, beware that there is a large body of changes to the Docker documentation coming via #261 for next release. This will likely need to be reworked against the new documentation once that's finalized.

PR #268, opened 2025-02-25
GUACAMOLE-2038: Document configuring outbound IPv6 connections from guacamole

pv2b:
This adds a documentation section for how to configure tomcat to prefer outbound IPv6 connections, which is useful in an IPv6-only environment.

NOTE: This PR should not be merged before #267 is merged, since there's a link to a chapter added in that PR, about configuring guacd for IPv6, which must be done when enabling this.

### DIFF
--- a/src/guacamole-docker.md
+++ b/src/guacamole-docker.md
@@ -182,8 +182,8 @@ not be able to start up, and you will see an error.
 
 :::{note}
 This section only applies to traffic originated by Gucamole Client,
-e.g. connections from Guacamole Client to guacd, or to authentication
-servers.
+e.g. connections from Guacamole Client to guacd, the database, or to
+authentication servers.
 
 Guacamole Client does not connect directly to RDP, VNC or SSH servers,
 this is done by guacd.


src/guacamole-docker.md
```diff
@@ -176,6 +176,31 @@ detail in the sections below. If the required configuration options for at
...
+A record is present in DNS. IPv6 will be used only if there are only AAAA
+records in DNS, or if an IPv6 address is hard-coded.
+
+Additionally, Guacamole will only attempt to make one connection to a single
+IP address, so if the IPv4 connection fails, it will not fall back to IPv6.
+This can be problematic for IPv6-only deployments, where such connections may
+fail with a "Network is unreachable" error.
+
+To override this, set the `JAVA_OPTS` environment variable to
+`-Djava.net.preferIPv6Addresses=true`.
```
mike-jumper, line 193: I don't think this is entirely correct:

* The JVM defaults to preferring IPv4. That's not to say this isn't worth documenting, but it's not something specific to the Guacamole webapp.
* As written, it sounds like connections to other machines (ie: remote desktops) will fail for IPv6, but this is not the case. The JVM configuration option in question controls only whether IPv6 is used by the JVM, which would affect only the connection to guacd itself and external auth systems (databases, LDAP, etc.).

  This is eventually clarified later, but I think we can avoid the confusion earlier by specifically noting what this affects. Users would need to set this option only in the event that they need IPv6 to communicate with their own guacd service, their own database, or other authentication backends.

mike-jumper (changes requested):

src/guacamole-docker.md
pv2b: I've re-worded this documentation section, hopefully this addresses your concerns

PR #269, opened 2025-02-25
GUACAMOLE-2039: Explain that other auth methods are possible in docker

pv2b:
Explicitly explain that other authentication methods, even such that are not explicitly documented as supported under Docker, are also possible to use, such as OpenID connect.

Technically, this doesn't add any new information since the "Configuring Guacamole when using Docker" section already explains using environment properties, but it does make this information easier to find.

### DIFF
--- a/src/guacamole-docker.md
+++ b/src/guacamole-docker.md
@@ -842,6 +842,21 @@ unusable if your authenticator app does not support setting these parameters.
 
 (guacamole-docker-history-recording-storage)=
 
+### Other authentication methods
+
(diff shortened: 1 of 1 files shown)


mike-jumper:
This is already on the way via the changes for [GUACAMOLE-374](https://issues.apache.org/jira/browse/GUACAMOLE-374), in particular #261 (which generalizes the documentation to cover Docker configuration alongside traditional native configuration).

mike-jumper:
Closing per above.

PR #283, opened 2026-02-14
GUACAMOLE-2085: Revise quickconnect extension documentation.

mike-jumper:
This PR builds off #274 and is intended to supersede those changes. The merge commits have been rebased away, and the em dashes have been replaced with normal dashes and spacing.

PR #274, opened 2025-06-19
GUACAMOLE-2085: Revise quickconnect extension documentation

flyn-org:
The primary goals of this revision are to:

	- Clarify guacamole.properties hosts the extension's configuration parameters.

	- Provide an example of a parameter list.

	- Provide more recommendations about how to configure the extension to meet security goals.

(I am currently awaiting a Jira account.)

PR #284, opened 2026-02-14
GUACAMOLE-2085: Revise quickconnect extension documentation.

mike-jumper:
OK, second try: this PR is identical to #283 except that my commit has been amended to correct the JIRA issue number.

From #283:

> This PR builds off #274 and is intended to supersede those changes. The merge commits have been rebased away, and the em dashes have been replaced with normal dashes and spacing.

### DIFF
--- a/src/adhoc-connections.md.j2
+++ b/src/adhoc-connections.md.j2
@@ -39,8 +39,8 @@ well-understood by administrators prior to implementing it:
 
   These and other factors should be taken into consideration when enabling
   this extension. You should combine your configuration of the parameters
-  allowed by this extension with features of the host system—such
-  as filesystem permissions, firewalls, and so on—to achieve your
+  allowed by this extension with features of the host system - such
+  as filesystem permissions, firewalls, and so on - to achieve your
   security needs.
 :::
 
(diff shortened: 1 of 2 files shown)


src/adhoc-connections.md.j2
```diff
@@ -22,17 +22,26 @@ well-understood by administrators prior to implementing it:
...
 
-* The extension provides users the ability not only to establish connections,
-  but also to set any of the parameters for a connection. There are security
-  implications for this - for example, RDP file sharing can be used to pass
-  through any directory available on the server running guacd to the remote
-  desktop. This should be taken into consideration when enabling this extension
-  and making sure that guacd is configured in a way that does not compromise
-  sensitive system files by allowing access to them.
+* This extension provides users the ability not only to establish connections,
+  but also set any of the parameters for a connection. This has security
```
necouchman: `but also set any` -> `but also to set any`

necouchman (changes requested):
One minor correction.

src/adhoc-connections.md.j2
mike-jumper: Sure - should be fixed now.

I also added emphasis around "this has security implications".

necouchman (approved):

PR #286, opened 2026-02-28
GUACAMOLE-2142: Clean up em-dash and line formatting on JSON authentication documentation.

necouchman:
@mike-jumper I took a stab at the clean-up, here - feel free to either suggest further changes or close and open a pull request if you have something else, already.

### DIFF
--- a/src/json-auth.md.j2
+++ b/src/json-auth.md.j2
@@ -282,26 +282,32 @@ Guacamole, will authenticate a user and grant them access to the connections
 described in the JSON (at least until the expires timestamp is reached, at
 which point the JSON will no longer be accepted).
 
-In summary, after obtaining the `base64` data of your encrypted, signed JSON, you can authenticate using one of the following methods:
+After obtaining the `base64` data of your encrypted, signed JSON, you need
+to URL-encode the data. There are a wide variety of ways to accomplish this
+with most programming languages providing mechanisms for URL-encoding data.
+The following is an example of a one-line Python command:
+```
+python3 -c "import urllib.parse; print(urllib.parse.quote('{your_base64_encoded_data}', safe=''))"
+```
 
-1. **URL Encoding Method**: Pass the URL-encoded base64 data directly in the URL, like so:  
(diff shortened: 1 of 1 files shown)


src/json-auth.md.j2
```diff
@@ -282,26 +282,34 @@ Guacamole, will authenticate a user and grant them access to the connections
...
-    Alternatively, you can use an online tool to encode it.
-2. **POST Request Method**: As shown in the `curl` example above, you can also make a POST request to `/api/tokens` with the `data` parameter in the request body, 
-    where the value is the raw base64-encoded encrypted JSON:
+    Many programming languages include tools for URL-encoding data, and there
+    are online tools that can provide it for you, as well.
+
+2. **POST Request Method**: As shown in the `curl` example above, you can also
+    make a POST request to the `api/tokens` endpoint of your Guacamole Client
+    install with the `data` parameter in the request body, where the value is
+    the raw base64-encoded encrypted JSON:
```
mike-jumper: Not newly introduced, but I think this is technically not correct. The request body for `api/tokens` is expected to be `application/x-www-form-urlencoded`.

necouchman: Fixed via rebase.

mike-jumper: Sorry, but the JSON part (also from before) is wrong, too, I think:

> ```json
> {
>     "method": "POST",
>     "data": "{your_url_encoded_data}"
> }
> ```

We don't use JSON for the auth endpoint, and none of our JSON uses a `method` parameter elsewhere.

necouchman: Got it - how's it look, now? Or is there a more proper way to represent `application/x-www-form-urlencoded` data?

mike-jumper: It should be essentially a query string, submitted as the request body. See: https://stackoverflow.com/a/4073451

necouchman: Okay, I've tweaked the formatting of that line.

mike-jumper (approved):

